PR #1, opened 2017-06-05
Run the game engine

gianluca-nitti:
For this first pull request I added the code to make this an actual engine facade. The engine runs on the main thread, while the Jetty web server runs on separate threads. Lots of "inspiration" and parts of code come from the standard PC facade. Here is how I set up my workspace, hoping it's the correct way (it should be according to the [wiki instructions](https://github.com/MovingBlocks/Terasology/wiki/Codebase-Structure#facades) for facades): 
- from my local clone of the [engine repository](https://github.com/MovingBlocks/Terasology), ran `gradlew fetchFacadeServer`
- linked my fork with the nested git repository

Put shortly, to test this branch you can follow this process:
- From the root of your local clone of [MovingBlocks/Terasology](https://github.com/MovingBlocks/Terasology), run `gradlew fetchFacadeServer`. On a freshly cloned copy I usually have no issues if I do a `gradlew build` before, while if I recall correctly I had the engine complaining about missing world generators when running without manually building the engine first. Maybe I didn't set up all the dependencies properly (*).
- then, move to the nested facade repository with `cd facades/Server` and reach this branch using the following commands:
```
git remote add fork "https://github.com/gianluca-nitti/FacadeServer"
git pull fork
git checkout engine-runner
```
- now you can go back to the root repository (`cd ../..`) and run this facade with `gradlew facades:Server:run`.

(*) about the Gradle configuration: I made some tweaks to the `build.gradle` according to the other facades, to setup a basic server configuration; it all works but I'm not sure if it's the correct way and how to correctly handle packaging/distribution.
Also, I added two command line arguments, one to set the engine data directory and the other for the game server port. I'm not sure if the web server port too should be settable via a command line switch, for the moment I didn't change the existing code which sets it according to the PORT environment variable.

Maybe we can also improve these logistics details later if necessary, I focused on setting up a working development environment since I'm a bit late on the timeline.

Running the project as a facade introduced some issues with paths, because when you run via `gradlew facades:Server:run` the working directory of the program is the root repository and not the root of the facade, so at the moment you get 404s if you try to access `localhost:8080/`; I fixed the path for the logs using the engine's `LoggingContext`, so they are correctly served at `localhost:8080/logs`. In my opinion we could put the static web files in an appropriate subdirectory of `src/main/resources/` but I think this will be relevant when I'll work on the frontend; my main goal at the moment is setting up the authentication API.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,8 @@ def jettyVersion = '9.3.8.v20160314'
 def jerseyVersion = '2.22.2'
 
 dependencies {
-    compile 'org.terasology.engine:engine:1.0.0'
+    //compile 'org.terasology.engine:engine:1.0.0'
+    compile project(':engine')
 
     compile group: 'org.eclipse.jetty', name: 'jetty-servlet', version: jettyVersion
     compile group: 'org.eclipse.jetty.websocket', name: 'websocket-server', version: jettyVersion
--- a/build.gradle
+++ b/build.gradle
@@ -99,3 +99,4 @@ run.dependsOn rootProject.moduleClasses
 run.dependsOn setupServerConfig
 run.dependsOn setupServerModules
 run.workingDir = rootDir
+run.args = ["-homedir=terasology-server"]
--- a/build.gradle
+++ b/build.gradle
@@ -2,24 +2,18 @@ plugins {
     id "org.standardout.versioneye" version "1.3.0"
 }
 
-apply plugin: 'java'
+apply from: "$rootDir/config/gradle/artifactory.gradle"
 apply plugin: 'application'
-apply plugin: 'eclipse'
-apply plugin: 'idea'
 
 apply from: 'config/gradle/versioning.gradle'
 
+import groovy.json.JsonBuilder
+
 mainClassName = "org.terasology.web.ServerMain"
-group = 'org.terasology.web'
 
-// Same repository configuration as root project
-repositories {
-   mavenCentral()
-
-    // MovingBlocks Artifactory instance for libs not readily available elsewhere plus our own libs
-    maven {
-        url "http://artifactory.terasology.org/artifactory/virtual-repo-live"
-    }
+ext {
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -30,7 +24,8 @@ def jettyVersion = '9.3.8.v20160314'
 def jerseyVersion = '2.22.2'
 
 dependencies {
-    compile 'org.terasology.engine:engine:1.0.0'
+    //compile 'org.terasology.engine:engine:1.0.0'
+    compile project(':engine')
```
skaldarnar, line 28: Does this use the local engine project instead of the artefact? If so, is this just for (local) testing or should it stay in?

skaldarnar (approved):
Tested it out and could start the server facade without problems (after some waiting time). I was able to access the log files at `localhost:8080/logs`.

skaldarnar:
@Cervator might want to have a quick look at the changes to `build.gradle`...

build.gradle
gianluca-nitti: Yes, it uses the local engine. I followed the [facade tutorial](https://github.com/MovingBlocks/Terasology/wiki/Codebase-Structure#facades) so that when you have the main repository (MovingBlocks/Terasology) as root and the nested repository at facades/Server, the engine in the root repository is used. The [PC facade is set up in the same way](https://github.com/MovingBlocks/Terasology/blob/develop/facades/PC/build.gradle#L41).

I'm not sure how packaging/deployment should be managed. Probably we should make a package with the engine and the facade as for the PC facade?

msteiger (approved):

Cervator:
Belated note about the `build.gradle` (feel free to ping me on Slack for anything important - I've got 4k unread GitHub notifications...): I'm fine with it here. Some bits are out of date but I've forked this to the test org Nanoware and pulled it locally to my Gradle redo workspace so I can include it at some point in the new stuff :-)

PR #2, opened 2017-06-13
Authentication management

gianluca-nitti:
# Overview

Here is the code which implements user authentication via HTTP and WebSocket as described in the [devlog updates](https://gianluca-nitti.github.io/GSoC-2017-devlog/) for weeks 2-3.

Here is a little summary of the updates (more details in the posts at the link above, and in the recent messages on Slack#server-api):
- This facade will provide access to various resources of a running Terasology server via HTTP and WebSocket clients, exchanging messages in the JSON data serialization protocol.
- The authentication protocol is designed to resemble the already existing one for regular Terasology client, to achieve a good security level. This means that an authentication handshake roughly happens in this way:
  - The client asks to authenticate
  - The server sends an hello message which contains it's public certificate along with some random data and a timestamp
  - The client **must** reply with another hello message composed in the same way (but using the client public certificate instead of the server one), followed by a verification payload which is obtained by concatenating the server hello (received) with the client hello (generated and sent) and RSA signing the result using the client private certificate as key.
  - The server verifies the signature, and if it's valid the client is authenticated.
  - On successful authentication, the server also sends a verification payload constructed in the same way but signed with the server private certificate
  - *Optionally*, the client can verify the server identity by verifying the signature against the server public certificate, and optionally disconnect if this verification fails and since the server is not who it claims to be
- For WebSocket, the handshake is simply a sequence of messages on the same socket; for HTTP, every "ping-pong" (e.g. "init->obtain server hello", "send client hello and signed data->receive server signed data or authentication failure") is a "request-response" pair; on the first request, the server will also send a session token as a response header and the client must include it as a request header for the other requests.

# How to test
There is a simple Javascript WebSocket client as a proof-of-concept that the RSA part can be handled by a browser (with the right libraries) [here](https://gist.github.com/gianluca-nitti/14e5d61d66d1f347330fa7f42f3db7da). For HTTP I tested manually using [Postman](https://www.getpostman.com/).

Note that the BigIntegers in the certificates are serialized as a base64 string representing their binary representation; **the [BigIntegerBase64Serializer](https://github.com/gianluca-nitti/Terasology/blob/09273389852b8573c3c0d2f1811f2dfb80d0dabd/engine/src/main/java/org/terasology/identity/storageServiceClient/BigIntegerBase64Serializer.java) class implemented in MovingBlocks/Terasology#2965, thus this branch will not compile if the root repository (where this facade is located as a nested repository in facades/Server) is not at that branch or that PR is merged**.
The reasons behind this is that handling very large numbers in JSON is unhandy, for example in Javascript external libraries must be used, and that by using base64 the messages are shorter (a little bandwidth is saved, but probably this isn't very relevant). This also means that directly pasting the "normal" configuartion file in the simple client linked above won't work, it needs to be re-encoded with the BigIntegers in the correct format. The hack I used was to start the regular client in debug mode with a breakpoint on the code that saves the configuration, and then using the IntelliJ debugger's "evaluate expression" feature to add the correct gson type adapter and obtaining a JSON string of the configuration with the numbers in the right format. Since the Java's BigInteger binary representation is simply [the two's complement representation of the number](https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html#toByteArray--) manually converting them with a tool [like this](https://conv.darkbyte.ru/) (pasting a number in the DEC field and getting the base64) should also work but I have not tried it yet.
Any suggestion is welcome, maybe another version of `BigIntegerBase64Serializer` in this repository should be made so that the server could handle both the representations?


src/main/java/org/terasology/web/authentication/AuthenticationFailedException.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.web.authentication;
+
+public class AuthenticationFailedException extends Exception {
+
+    public AuthenticationFailedException(boolean invalidCertificate) {
+        super(invalidCertificate ? "Received invalid client certificate" : "Received invalid verification signature");
```
msteiger: I'd rather support a `String message` parameter instead of this flag like many other Exception classes do.

src/main/java/org/terasology/web/client/HeadlessClient.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public void send(Event event, EntityRef target) {
+
+    }
+
+    @Override
+    public ViewDistance getViewDistance() {
+        return null;
```
msteiger: I'd rather use a ViewDistance with range 0 instead of returning `null`. This could fail at some point of time (leading to a NPE), even if it works at the first sight.

Probably a new ViewDistance enum needs to be created for this. We could use LEGALLY_BLIND until then.

src/main/java/org/terasology/web/io/GsonMessageBodyHandler.java
```diff
@@ -53,6 +53,7 @@
 public final class GsonMessageBodyHandler implements MessageBodyWriter<Object>, MessageBodyReader<Object> {
 
     private static final Gson GSON = new GsonBuilder()
+        .disableHtmlEscaping()
```
msteiger, line 56: Nice!

src/main/java/org/terasology/web/io/JsonSession.java
```diff
@@ -0,0 +1,92 @@
...
+import org.terasology.web.authentication.HandshakeHello;
+import org.terasology.web.client.HeadlessClient;
+import org.terasology.web.client.HeadlessClientFactory;
+
+import java.math.BigInteger;
+
+public class JsonSession {
+
+    private static final Gson GSON = new GsonBuilder()
+            .registerTypeHierarchyAdapter(BigInteger.class, BigIntegerBase64Serializer.getInstance())
```
msteiger: I would just use registerTypeAdapter. No need to deal with subclasses of `BigInteger`. Just a personal opinion, though. You can, of course, just leave it, since it's perfectly valid.

src/main/java/org/terasology/web/io/JsonSession.java
```diff
@@ -0,0 +1,92 @@
...
+        if (isAuthenticated()) {
+            return new ActionResult(ActionResult.Status.UNAUTHORIZED, "Already authenticated");
+        }
+        try {
+            ClientAuthenticationMessage clientAuthentication = GSON.fromJson(clientMessage, ClientAuthenticationMessage.class);
+            byte[] serverVerification = authHandler.authenticate(clientAuthentication);
+            String clientId = clientAuthentication.getClientHello().getCertificate().getId();
+            client = headlessClientFactory.connectNewHeadlessClient(clientId);
+            return new ActionResult(GSON.toJsonTree(serverVerification));
+        } catch (NullPointerException | JsonSyntaxException ex) {
```
msteiger, line 80: I would try to avoid catching NPEs .. is this a leftover from a debug session?

src/main/java/org/terasology/web/servlet/HttpAPIServlet.java
```diff
@@ -0,0 +1,102 @@
...
+import javax.ws.rs.core.Response;
+import java.util.AbstractMap;
+import java.util.Map;
+import java.util.UUID;
+
+@Path("http")
+public class HttpAPIServlet {
+
+    private static final String SESSION_TOKEN_HEADER = "Session-Token";
+    private final Map<String, JsonSession> sessions = Maps.newHashMap(); //maps session tokens with the active sessions
```
msteiger: Just a comment: you can just use new `HashMap<>()` - with Java 8 this helper method is no longer that useful.

src/main/java/org/terasology/web/servlet/HttpAPIServlet.java
```diff
@@ -0,0 +1,102 @@
...
+    private JsonSession anonymousSession;
+
+    // TODO: call this after engine initialization
+    public void initAnonymourSession() {
```
msteiger: Typo?

src/main/java/org/terasology/web/servlet/HttpAPIServlet.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    private JsonSession getSession(HttpServletRequest request) {
+        String token = request.getHeader(SESSION_TOKEN_HEADER);
+        if (token == null) {
+            return anonymousSession;
+        }
+        JsonSession session = sessions.get(token);
+        if (session == null) {
+            throw new JsonWebApplicationException("Invalid session token", Response.Status.FORBIDDEN); //non-existing token token -> forbidden
```
msteiger: Duplicate word? I try to stick with the 120 chars per line limit, but that's me being pedantic ;-)

src/main/java/org/terasology/web/servlet/HttpAPIServlet.java
```diff
@@ -0,0 +1,102 @@
...
+    private static final String SESSION_TOKEN_HEADER = "Session-Token";
+    private final Map<String, JsonSession> sessions = Maps.newHashMap(); //maps session tokens with the active sessions
+    private JsonSession anonymousSession;
+
+    // TODO: call this after engine initialization
+    public void initAnonymourSession() {
+         anonymousSession = new JsonSession();
+    }
+
+    private Map.Entry<String, JsonSession> newSession() {
```
msteiger: Working with Map's more or less internal classes doesn't look like a proper way to return Key-Value-Pairs to me. In this particular case, I'd just inline this method. It's fairly trivial and called only once. Maybe a comment "new session" helps to clarify the intent.

src/main/java/org/terasology/web/servlet/HttpAPIServlet.java
```diff
@@ -0,0 +1,102 @@
...
+        return session.finishAuthentication(data);
+    }
+
+    @DELETE
+    @Path("auth")
+    @Produces(MediaType.APPLICATION_JSON)
+    public ActionResult logout(@Context HttpServletRequest request) {
+        String token = request.getHeader(SESSION_TOKEN_HEADER);
+        if (!sessions.containsKey(token)) {
+            throw new JsonWebApplicationException("Invalid session token", Response.Status.FORBIDDEN);
```
msteiger: Alternatively, you could just return "404 - Not Found"

src/main/java/org/terasology/web/servlet/JsonWebApplicationException.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.web.servlet;
+
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+public class JsonWebApplicationException extends WebApplicationException {
```
msteiger, line 22: I think that it was possible to map the exception to different HTTP response codes. This could be useful.

src/main/java/org/terasology/web/webSocket/WsHandler.java
```diff
@@ -14,37 +14,67 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.terasology.web.io.ActionResult;
+import org.terasology.web.io.JsonSession;
 
 /**
  * Manages one websocket session
  */
-public class EventSocket extends WebSocketAdapter {
+public class WsHandler extends WebSocketAdapter {
```
msteiger: How about `WebsocketHandler`?

src/main/java/org/terasology/web/webSocket/WsHandler.java
```diff
@@ -14,37 +14,67 @@
...
 
-    private static final Logger logger = LoggerFactory.getLogger(EventSocket.class);
+    private static final Logger logger = LoggerFactory.getLogger(WsHandler.class);
+    private static final Gson GSON = new GsonBuilder().disableHtmlEscaping().create();
```
msteiger: Would it make sense to have one common GSON instance that is used in both cases?

src/main/java/org/terasology/web/webSocket/ClientMessage.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public MessageType getMessageType() {
+        return messageType;
+    }
+
+    public JsonElement getData() {
+        return data;
+    }
+
+    public void checkValid() throws InvalidClientMessageException {
```
msteiger, line 39: I think that it would be better to perform the sanity checks in the constructor instead of hoping that the caller will invoke this method.

You can do that by subclassing [InstanceCreator](https://static.javadoc.io/com.google.code.gson/gson/2.6.2/com/google/gson/InstanceCreator.html) and registering it.

src/main/java/org/terasology/web/webSocket/WsHandler.java
```diff
@@ -14,37 +14,67 @@
...
-        trySend("Connected!");
+        jsonSession = new JsonSession();
     }
 
     @Override
     public void onWebSocketText(String message) {
         super.onWebSocketText(message);
+        ClientMessage deserializedMessage;
+        try {
+            deserializedMessage = GSON.fromJson(message, ClientMessage.class);
```
msteiger: I think that this method would be easier to read if the variable was declared inside the try/catch block and the switch statement below would be in a separate method.

src/test/java/org/terasology/web/io/JsonSessionTest.java
```diff
@@ -0,0 +1,141 @@
...
+            Preconditions.checkNotNull(authenticationMessage.getClientHello());
+            Preconditions.checkNotNull(authenticationMessage.getSignature());
+            if (!nextResult) {
+                throw new AuthenticationFailedException(true);
+            }
+            return new byte[]{0, 0, 0};
+        }
+    }
+
+    private static final Gson GSON = new GsonBuilder()
```
msteiger, line 72: Maybe another reason to create a shared instance of GSON for the HTTP/WS marshalling.

msteiger (changes requested):
Great work, @gianluca-nitti - just a few minor things.

src/main/java/org/terasology/web/io/JsonSession.java
gianluca-nitti: I agree that the Hierarchy is useless here since there are no subclasses and I can't remember why I used it, most likely misclicked an IDE autocompletion suggestion.

src/main/java/org/terasology/web/io/JsonSession.java
gianluca-nitti: Not a debug leftover, the reason it's there is that an invalid JSON message sent by the client could miss some field (for example the signature) and thus the deserialized object will have those fields set to `null`, resulting in NPEs during the verification process.
Anyway, I agree that catching unchecked exception is not that good. I think I'll fix this by ~~adding a validation method~~ properly subclass InstanceCreator to throw a custom exception if some mandatory fields are null, like I did for the deserialization of WebSocket resource request messages.

src/main/java/org/terasology/web/webSocket/ClientMessage.java
gianluca-nitti: I agree that relying on the caller to perform checks isn't very good, however at the moment I can't find any real alternative. I read the InstanceCreator documentation and as far as I understand it allows you to define code that will run **before** the deserialization and must return an instance, which fields will then be overwritten with the deserialized data. So, in the [T createInstance(Type type)](https://static.javadoc.io/com.google.code.gson/gson/2.6.2/com/google/gson/InstanceCreator.html#createInstance-java.lang.reflect.Type-) method I don't have access to the data and thus can't validate it.

I also considered using a custom [deserializer](https://static.javadoc.io/com.google.code.gson/gson/2.6.2/com/google/gson/JsonDeserializer.html), but I'm not sure it would be of much help:
- I should find a way of deserializing the data anyway; using context.deserialize() would cause an infinite loop because it would call the same `deserialize` method, so the remaining options are manual deserialization (probably not that hard, but would defeat the purpose of using gson I guess?) or creating a new GSON instance (losing the context, i.e. registered adapters; in this case it shouldn't be a big deal since there are just a `String` field and a `JsonElement` one, however the other class where I should perform similar checks to avoid the NullReferenceException catch block we are talking about in [the other comment thread](https://github.com/MovingBlocks/FacadeServer/pull/2#discussion_r122599388) - `ClientAuthenticationMessage` - would be much more troubling because the structure contains `BigInteger` and `byte[]` fields, which have custom deserializers).
- I couldn't throw a custom exception such as `InvalidClientMessageException` because of the signature of the method in the interface that only allows to throw `JsonParseException`. This probably isn't much of a problem though since I could throw it as the cause of the JsonParseException.

src/main/java/org/terasology/web/webSocket/WsHandler.java
gianluca-nitti: You mean shared between `WebSocketHandler` and `GsonMessageBodyHandler`? Not sure about that since they serve different purposes.
- The instance in `WebSocketHandler` is used for "container" JSON objects, like `ClientMessage`, which encapsulate objects to send to `JsonSession` for further processing and add some fields specific to the WebSocket (like messageType, which in HTTP is not necessary because the information is in the URL).
- The instance in `GsonMessageBodyHandler` is used for incoming and outgoing HTTP messages; has the custom type adapters for Version and Name that in my understanding shouldn't be necessary for WebSocket (but maybe aren't necessary here anymore, too)?

src/main/java/org/terasology/web/servlet/JsonWebApplicationException.java
gianluca-nitti: I'm not sure I get what you mean. A `JsonWebApplicationException` is constructed by specifying a status code and a message, which is sent as the HTTP response code, which is sent in a JSON object like `{"message": "the error message"}`. They are provided by the code which builds the exception.

I can't figure out a better way but probably it's just me, if you think there is one could you please be more specific?

gianluca-nitti:
Thank you for the very in-depth review.

I fixed most of the issues and left some comments on the ones I'm not sure how to improve.

Summing up:
- For the global gson instance: as you can see in the previous comments I'm not sure it's the best solution since they are used for different purposes, but it's also true that there aren't conflicting adapters. Anyway, do you think having a class keeping the instance with all the various adapter in a static field with a public getter to be a good solution? I think this could be a good compromise (or maybe exactly what you meant by shared instance)?
- The other issue is the sanity checks for nulls on the deserialized objects sent as client-to-server messages. I'm thinking that a quick solution could be:
  - adding a custom deserializer for `ClientMessage` (in `websocket` package) which deserializes using a separate gson instance (as I said before, in my understanding, using the shared one would result in a loop), then performs the checks and throws a `JsonParseException` with an `InvalidClientMessageException` as cause if the message is not valid, and
  - Performing the checks in `AuthenticationHandshakeHandlerImpl` before doing the operations that could throw an NPE and throw a custom exception instead, since writing a validating deserializer for `ClientAuthenticationMessage` would be tricky due to the required custom type adapters as discussed before.

What do you think about it?

src/main/java/org/terasology/web/webSocket/WsHandler.java
```diff
@@ -14,37 +14,67 @@
...
+            deserializedMessage.checkValid();
+        } catch (JsonSyntaxException ex) {
+            trySendResult(ActionResult.JSON_PARSE_ERROR);
+            return;
+        } catch (NullPointerException | InvalidClientMessageException ex) {
```
msteiger: I think that no NPE can happen here.

msteiger (approved):

PR #3, opened 2017-06-22
Resource management

gianluca-nitti:
This PR contains the code which allows the HTTP and WebSocket servers to expose resources that can be accessed by clients. Also, two resources are included: `ConsoleResource` allows an authenticated user to execute console commands, and emits events when messages are written to the console, and `OnlinePlayersResource` can be queried (and notifies it's changes) to obtain a list of the names of the players currently connected to the server.
An overview of the implemented class structure is aviable on the [devlog site](https://gianluca-nitti.github.io/GSoC-2017-devlog/), in the 18 June (week 3) update.

A better error handling system for invalid request messages has been implemented too. As recently suggested by @msteiger, I implemented a custom gson type adapter factory (`ValidatorTypeAdapterFactory`) which delegates the deserialization work to the default deserializer for a given data type, and then if the obtained object's type implements the `Validable` interface, its `validate` method is automatically called as part of the deserialization process so that if the data sent by the client is not consistent (i.e. mandatory fields are missing or similar issues) the deserialization fails. Now the classes which encapsulate data deserialized from client messages still have a `validate` method, but the usage is different from the old `checkValid` methods that were called manually after deserialization: the methods are there because the message classes implement `Validable`, but the code that calls `gson.fromJson(...)` doesn't need to call them because the validation is intrinsically performed during deserialization when the deserializer detects that the class implements `Validable`.


src/main/java/org/terasology/web/resources/ConsoleResource.java
```diff
@@ -0,0 +1,54 @@
...
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.entitySystem.event.ReceiveEvent;
+import org.terasology.entitySystem.systems.RegisterSystem;
+import org.terasology.logic.console.Console;
+import org.terasology.logic.console.Message;
+import org.terasology.logic.console.MessageEvent;
+import org.terasology.network.ClientComponent;
+
+@RegisterSystem
+public class ConsoleResource extends EventEmittingResource<Message> implements DefaultComponentSystem, WritableResource<String> {
```
msteiger: Just a general note: if possible, I use other components as fields instead of inheriting from them. See https://en.wikipedia.org/wiki/Composition_over_inheritance for details - do you think it's possible here?

src/main/java/org/terasology/web/resources/OnlinePlayersResource.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+@RegisterSystem
+public class OnlinePlayersResource extends ObservableReadableResource<List<String>> implements DefaultComponentSystem {
+
+    private NetworkSystem networkSystem;
+
+    OnlinePlayersResource(NetworkSystem networkSystem) {
```
msteiger: Why not just inject the NetworkSystem with `@In` here?

src/main/java/org/terasology/web/resources/ResourceManager.java
```diff
@@ -0,0 +1,87 @@
...
+            putResource(new OnlinePlayersResource(context.get(NetworkSystem.class)));
+        }
+        registerEventHandlers(context);
+    }
+
+    private void registerEventHandlers(Context context) {
+        EventSystem eventSystem = context.get(EntityManager.class).getEventSystem();
+        for (Map.Entry<String, Resource> entry: resources.entrySet()) {
+            Resource resource = entry.getValue();
+            if (resource instanceof ComponentSystem) {
```
msteiger: If I remember correctly, all systems that are annotated correctly should receive events. Am I missing something here?

src/main/java/org/terasology/web/resources/ResourceManager.java
```diff
@@ -0,0 +1,87 @@
...
+        if (resource == null) {
+            throw new ResourceAccessException(new ActionResult(ActionResult.Status.NOT_FOUND, "Resource not found."));
+        }
+        if (type.isAssignableFrom(resource.getClass())) {
+            return type.cast(resource);
+        }
+        throw new ResourceAccessException(new ActionResult(ActionResult.Status.ACTION_NOT_ALLOWED, "This resource does not support the requested action."));
+    }
+
+    public <T extends Resource> Set<T> getAllAs(Class<T> type) {
```
msteiger: The name might be misleading, since the methods just skip over entries that do not match. How about `getAll(Class<T>)`?

src/main/java/org/terasology/web/servlet/HttpAPIServlet.java
```diff
@@ -26,37 +26,45 @@
...
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
 @Path("http")
```
msteiger: I know that this wasn't touched in this PR, but I'd rather use something like "api" for the root path.

src/main/java/org/terasology/web/webSocket/ClientToServerMessage.java
```diff
@@ -36,13 +38,14 @@ public JsonElement getData() {
         return data;
     }
 
-    public void checkValid() throws InvalidClientMessageException {
+    @Override
+    public void validate() throws InvalidClientMessageException {
```
msteiger, line 42: Good!

src/main/java/org/terasology/web/webSocket/WebSocketHandler.java
```diff
@@ -72,28 +78,50 @@ public void onWebSocketError(Throwable cause) {
...
+
+    private void handleResourceRequest(JsonElement requestMessage) {
+        ResourceRequestClientMessage deserializedMessage;
+        try {
+            deserializedMessage = GSON.fromJson(requestMessage, ResourceRequestClientMessage.class);
+        } catch (JsonSyntaxException ex) {
+            trySendResult(new ActionResult(ex));
+            return;
+        }
+        String resourceName = deserializedMessage.getResourceName();
```
msteiger: I would put this part inside the `try { .. }` or in a separate method that is called from the try block. This makes it easier to read.

src/test/java/org/terasology/web/io/JsonSessionTest.java
```diff
@@ -15,11 +15,13 @@
  */
 package org.terasology.web.io;
```
msteiger, line 16: Very nice test setup!

src/main/java/org/terasology/web/resources/OnlinePlayersResource.java
gianluca-nitti: Initially I used constructor injection since it looked more test-friendly but I now (9be5efa commit) switched to the `@In`s to make it more coherent with the various other classes I saw in the engine repository, and after all it just requires a few more lines in the test cases, definitely not a big deal. Thanks for the suggestion.

src/main/java/org/terasology/web/resources/ResourceManager.java
gianluca-nitti: As far as I understand, it's completely automatic only if they are in a Module. The ones in the engine repository (like for example `ChatSystem`) are registered because they are part of the `engine` module, however I can't understand who "decides" whether a class on the classpath of the running JVM is part of it; I tried but my classes in FacadeServer are not automatically registered. Maybe I should investigate further and find out how to have them automatically registered if possible.

I changed `ResourceManager` (commit 	4a77c50) to use `ComponentSystemManager` to *almost* automatically subscribe to events and inject required context entries in each resource which is a `ComponentSystem`. I think it's a bit better than before since now I'm not directly interacting with `EventSystem` and `InjectionHelper` but delegating the work to `ComponentSystemManager` which in my understanding is more high-level and is designed exactly for this purpose. However it still isn't *fully* automatic as in Modules, is it okay for you?

src/main/java/org/terasology/web/resources/ConsoleResource.java
gianluca-nitti: Okay, I'm reading about the pattern and I'll keep it in mind for the future. However in this particular case inheritance fits better:
- In my understanding, since it needs to receive events, the class must be a `ComponentSystem` in some way; `DefaultComponentSystem` just extends `ComponentSystem` by providing do-nothing default implementations for the various preInitialize, initialize, etc methods.
- `WritableResource` provides a common interface for any resource that accepts write operations.
- `EventEmittingResource` works like `java.util.Observable`.

Maybe you already noticed but just in case, [here is the class diagram](https://gianluca-nitti.github.io/GSoC-2017-devlog/images/ResourceClassDiagram.PNG) for resource management if you are interested. I'm not sure how it could be refactored using composition, to me it looks okay since it's mostly a way to mark the various type of resources.

PR #4, opened 2017-07-18
Games resource

gianluca-nitti:
## Contents
The code on this branch adds support for management of the savegames stored on the server from the web client/API, as described in the 16 July (week 7) [devlog update](https://gianluca-nitti.github.io/GSoC-2017-devlog/).

## Before merging
I'm marking this as WIP since there are a couple of parts that I think need to be improved before this branch is merged:
- ~~The `getLoggingPhase()` method of the `StateLoading` and `StateInGame` engine states is used to determine the name of the game which is running or loading, which, as pointed out in the comments in the code, works but in my opinion is not a very good idea, because for example the `getLoggingPhase()` might be changed in the future to provide additional information. So I'm considering opening a pull request in [MovingBlocks/Terasology](https://github.com/movingblocks/terasology) which adds a more suitable method to get this information (e.g. `getGameTitle()` or something like that).~~ Fixed in 	1a087c4 by using `currentState.getContext().get(Game.class).getName();`

- ~~As pointed out in the blog update `AvailableModulesResource` is an hack to circumvent the fact that CORS is not enabled on the [Meta Server](https://github.com/MovingBlocks/meta-server),thus the list of available modules can't be obtained directly from a web browser, so the game server is being used as a proxy. As described in the blog post we could either enable CORS on the meta server or improve `AvailableModulesResource` to provide different or additional information, for example only listing the modules installed on the game server.~~ `AvailableModulesResource` now provides information about **installed** modules and world generators.
- ~~When creating new games, any module can be specified from the web frontend, but if a module is not installed on the game server the dependency resolution fails. This is also the reason I still haven't implemented the feature to modify an existing savegame's module list (actually existing savegames can be deleted, backed up or renamed). Two possible solutions could be:~~
    - ~~Download the missing modules on-the-fly when the game is created.~~
Picked this solution in commit 2e15c95:
    - Show only the installed modules on the game creation screen, and add a separate resource class (and a new tab in the frontend application) to install new modules in the server.

Also, as pointed out at the end of the blog post, the engine can be started in this new state (`StateEngineIdle`) suggested by @msteiger by specifying the `-dontStartDefault` command line switch, but at the moment there is no way to actually avoid the creation of a default game because you can't login to the web interface without a valid client identity certificate, which is generated when connecting to a running game. The planned solution to this problem is to set up a sort of admin password that can be used to login to the web interface to only administer savegames without a client identity (but obviousluy as always any idea is welcome); let me know if I should add it on this branch or wait for this pull request to be reviewed and add it later.

## How to test
It's possible to access the new resources through the "Games and Modules" tab in the web client, you can either use the version hosted on GitHub Pages [here](https://gianluca-nitti.github.io/FacadeServer-frontend/) or clone the [repository](https://github.com/gianluca-nitti/facadeserver-frontend) locally and following the build instructions in the README (basically, you need to have `npm` installed, install the dependencies including the development ones and run `npm run web`). I'm aware that there is still some small usability problem in the frontend application but it shouldn't be a problem when using it for testing the server. ~~For example, if you connect to an not existing server, no warning is shown and various buttons become unresponsive (as pointed out by @skaldarnar last week). This week I mainly focused on the server and I'm going to fix this soon, in the meanwhile please press the "Connect" button on the "Server Address" dialog only when the local server is running and has completed initialization.~~ Fixed.


src/main/java/org/terasology/web/resources/games/AbstractAction.java
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.web.resources.games;
+
+import org.terasology.engine.module.ModuleManager;
+import org.terasology.engine.paths.PathManager;
+import org.terasology.web.resources.ResourceAccessException;
+
+public abstract class AbstractAction {
```
msteiger: I always try to make an abstract class implement an interface, to allow implementors to provide alternative content. Since this class just provides trivial code, it could be moved into the actual implementation and converted directly into an interface. But this would require removing the `setPathManager` method for the tests. Hmm .. tricky.

src/main/java/org/terasology/web/resources/games/BackupGameAction.java
```diff
@@ -0,0 +1,55 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class BackupGameAction extends AbstractExistingGameAction {
+
+    @Override
+    public void perform(String gameName) throws ResourceAccessException {
+        String backupName = gameName + "_backup_" + LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyy_MM_dd_HH_mm_ss"));
```
msteiger: The formatter could be a constant

src/main/java/org/terasology/web/resources/games/NewGameAction.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        WorldInfo worldInfo = new WorldInfo(TerasologyConstants.MAIN_WORLD, newGameManifest.getSeed(),
+                (long) (WorldTime.DAY_LENGTH * 0.025f), worldGenerator);
+        newGameManifest.addWorld(worldInfo);
+
+        return newGameManifest;
+    }
+
+    void setFields(String fGameName, String fSeed, List<Name> fModules, SimpleUri fWorldGenerator) {
```
msteiger: Looks like something that is useful for testing. Would it make sense to convert this into a package private constructor?

src/main/java/org/terasology/web/resources/games/NewGameAction.java
```diff
@@ -0,0 +1,101 @@
...
+        ResolutionResult result = dependencyResolver.resolve(modules);
+        if (!result.isSuccess()) {
+            throw new ResourceAccessException(new ActionResult(ActionResult.Status.GENERIC_ERROR, "Failed to resolve all module dependencies"));
+        }
+        for (Module m: result.getModules()) {
+            newGameManifest.addModule(m.getId(), m.getVersion());
+        }
+
+        WorldInfo worldInfo = new WorldInfo(TerasologyConstants.MAIN_WORLD, newGameManifest.getSeed(),
+                (long) (WorldTime.DAY_LENGTH * 0.025f), worldGenerator);
```
msteiger, line 97: This looks like copy&paste from some other place. I suspect that it wasn't possible to reuse the code. This world time setting is something for a different PR, I guess.

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
msteiger, line 1: What is this?

msteiger (approved):
Very nice, good work @gianluca-nitti

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
gianluca-nitti: It enables an incubating feature of mockito to mock final classes, which is done in some of the tests: https://github.com/mockito/mockito/wiki/What's-new-in-Mockito-2#mock-the-unmockable-opt-in-mocking-of-final-classesmethods

src/main/java/org/terasology/web/resources/games/NewGameAction.java
gianluca-nitti: Yes, these 2 lines are taken from [StateHeadlessSetup](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/subsystem/headless/mode/StateHeadlessSetup.java#L141); the time offset is set to that value to avoid the player spawning in a completely dark environment. In my understanding though it's not possible to reuse the whole [createGameManifest](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/subsystem/headless/mode/StateHeadlessSetup.java#L109) method since it just uses the default settings for title, seed and world generator while here we are using user-supplied values.

As you suggest, in another branch I could implement a Resource to allow to view and/or manipulate the world time of an existing savegame or maybe the currently running game. In my understanding I still need to set the world time to some value when initializing the game manifest, I could set it to zero if you prefer?

src/main/java/org/terasology/web/resources/games/AbstractAction.java
gianluca-nitti: Okay, refactored the class structure a bit and converted this to an interface in latest commit (373bbfd).

src/main/java/org/terasology/web/resources/games/NewGameAction.java
gianluca-nitti: Converted to a constructor as part of commit 373bbfd.

src/main/java/org/terasology/web/resources/games/NewGameAction.java
msteiger: No, it's fine. Sorry for distracting you. This world-time setting is set at various places and I'd like to centralize at some time, but that's a different story.

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
msteiger: Ah, that's good to know. Thanks!

msteiger:
Thank you very much!

PR #5, opened 2017-08-06
Modules install resource

gianluca-nitti:
This small pull request adds the feature which allows server administrators to install modules in the server using the web interface or API.

The added resource, `ModuleInstallerResource`:
- implements `WritableResource<Name[]>` in the sense that you can write to it (i.e. with a POST HTTP request or with the appropriate WebSocket message) a list of modules you want to install (which does not need to include all the dependencies; the dependency resolution happens server-side and the server adds the required dependencies - if any - to the download list automatically)
- extends `ObservableReadableResource<String>` in the sense that you can read from it the current status of the installation, and every connected WebSocket client is automatically notified when the status changes (i.e. a progress in the download is made, or the installation has been completed).

**Important**: for this to work, the engine must be built including my commits on pull request MovingBlocks/Terasology#3028; FacadeServer won't compile if classes introduced in that PR, like `ModuleInstallManager`, are missing.

### DIFF
--- a/src/main/java/org/terasology/web/resources/ResourceManager.java
+++ b/src/main/java/org/terasology/web/resources/ResourceManager.java
@@ -25,6 +25,7 @@ import org.terasology.registry.InjectionHelper;
 import org.terasology.web.io.ActionResult;
 import org.terasology.web.resources.games.GamesResource;
 import org.terasology.web.resources.modules.AvailableModulesResource;
+import org.terasology.web.resources.modules.ModuleInstallerResource;
 
 import java.util.HashMap;
 import java.util.HashSet;
@@ -52,6 +53,7 @@ public final class ResourceManager {
         registerAndPutResource(context, new EngineStateResource());
         registerAndPutResource(context, new GamesResource());
         registerAndPutResource(context, new AvailableModulesResource());
+        registerAndPutResource(context, new ModuleInstallerResource());
         if (gameState instanceof StateIngame) {
             registerAndPutResource(context, new ConsoleResource());
             registerAndPutResource(context, new OnlinePlayersResource());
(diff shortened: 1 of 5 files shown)


gianluca-nitti:
Update: while doing some additional testing I noticed that the WebSocket connection between the server and the web interface was frequently dropping when using this feature. After researching a bit of information about the exception that was being thrown by some Jetty internal class, I found out that there was a concurrency issue - multiple threads sometimes were trying to send WebSocket messages at the same time (when the `write` method of a `WritableResource` returns, an `ActionResult` whose content depends on whether the method threw a `ResourceAccessException` or not is sent to the client, but inside that method `ModuleInstallerResource` starts a parallel task to download the new modules which can send messages to report progress), and the previously used method didn't support this.
So I solved this issue in the latest commit by simply switching to another override of the `sendString` method of the `RemoteEndpoint` Jetty class; the currently used version is non-blocking, returns immediately and then calls a callback after the message has been sent. Looks like everything works correctly now. The method's [javadoc](http://www.eclipse.org/jetty/javadoc/9.4.4.v20170414/org/eclipse/jetty/websocket/api/RemoteEndpoint.html#sendString-java.lang.String-org.eclipse.jetty.websocket.api.WriteCallback-) isn't very exhaustive but the [comment](https://bugs.eclipse.org/bugs/show_bug.cgi?id=474488#c1) in the Jetty bug tracker talks about an internal buffer so in my understanding the messages are queued and then sent once at a time to the underlying socket.

msteiger:
Great find about the Jetty "deficiency".

msteiger (approved):

Cervator:
Engine PR has been merged, I think this one can be as well now :-)

msteiger:
Thanks @Cervator and @gianluca-nitti of course.

PR #6, opened 2017-08-06
Add resources to obtain or modify server's MOTD and port

gianluca-nitti:
This code adds read and write support for two server configuration entries (the only ones meaningful for a server which I could find, actually): the network listen port and the Message Of The Day (MOTD).

The read access is public (everyone which connects to the server via HTTP or WebSocket can see the value of this settings without authenticating), while the write access requires authentication and requires the client's ID to be in the server admin list (which is, at the moment, stored in a JSON file containing an array of client IDs which has to be manually edited on the server and is read by [this class](https://github.com/MovingBlocks/FacadeServer/blob/2f932ced0876c31d99be3988c955763f81dbedb5/src/main/java/org/terasology/web/ServerAdminsManager.java); it's in the process of being improved for better management, allowing public access until a first client identity is generated, and allowing an admin to add other admins - but anyway this is another story for another pull request).

The [frontend application](https://github.com/gianluca-nitti/FacadeServer-frontend) already supports this feature. In the home tab you can see these infos (if the server is providing them) and from the Settings tab they can be edited (an error message will be shown if you try to save the changes without being authenticated, or if you are authenticated but not in the admin list).

### DIFF
--- a/src/main/java/org/terasology/web/resources/ResourceManager.java
+++ b/src/main/java/org/terasology/web/resources/ResourceManager.java
@@ -23,6 +23,8 @@ import org.terasology.engine.modes.StateIngame;
 import org.terasology.entitySystem.systems.ComponentSystem;
 import org.terasology.registry.InjectionHelper;
 import org.terasology.web.io.ActionResult;
+import org.terasology.web.resources.config.ServerMotdResource;
+import org.terasology.web.resources.config.ServerPortResource;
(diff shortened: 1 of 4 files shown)


msteiger (commented):
Very nice and clean!

gianluca-nitti:
Just fixed the conflict caused by the merge of #5 in e6601cc.

msteiger:
Thanks!

PR #8, opened 2017-08-16
Manage administrative permissions

gianluca-nitti:
This pull request contains the code to manage a list of client IDs recognized as server administrators, as described in the first part of the August 13 (week 11) [blog post](https://gianluca-nitti.github.io/GSoC-2017-devlog/).

Short recap:
- Being a server administrators means having write access to the module installer, configuration and savegame list.
- When a new server (i.e. empty data directory) is started for the first time, access to admin-restricted resources is public. The reason for this is that when the server is in the idle state (not running any game) it's not possible to connect using the regular game client, which is what triggers the generation of a client identity; it's only possible to connect anonymously using the web client or API.
- As soon as a regular client connects, that client becomes the only server administrator. the recommended workflow for setting up a server is to start it, access the web interface and create and start a game, then connect using the game client to restrict the administrative access to yourself.
- Using the web interface, an authenticated administrator can add or remove other client IDs from the administrator list by either:
    - pasting the client ID of the new admin, or
    - if a game is running (thus there is a 1:1 correspondence between nicknames and client IDs), picking a name from the list of connected players.
- If an administrator removes everyone from the list including himself so that the admin list becomes empty again, the access to admin-restricted resources becomes public again.

A clarification about the two methods added to `WritableResource`:
- `boolean writeRequiresAuthentication()` must return true if the resource is only writable by a **non-anonymous** player, i.e. it interacts with the in-game entities of the running game and thus a client entity is required to access it. For example to write to `ConsoleResource` (execute a command) you need a client entity; to write to `ModuleInstallerResource` (installing modules) you don't, because you are not manipulating in-game entities.
- `boolean writeIsAdminRestricted()` must return true if writing to the resource is restricted to server administrators; it's possible that a resource is restricted to administrator because it modifies server-wide (and not game-local) files or configuration, but isn't restricted to authenticated clients because it doesn't manipulate in-game objects. In this case, when the admin list is empty (i.e. early server setup phase) this resource can be written to by everyone without authentication.

### DIFF
--- a/src/main/java/org/terasology/web/resources/OnlinePlayerMetadata.java
+++ b/src/main/java/org/terasology/web/resources/OnlinePlayerMetadata.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.web.resources;
+
+import org.terasology.network.Client;
+import org.terasology.rendering.nui.Color;
+
+import java.util.Objects;
+
+public class OnlinePlayerMetadata {
+
+    private String id;
+    private String name;
+    private RgbaColor color;
+
+    public OnlinePlayerMetadata(String id, String name, Color color) {
+        this.id = id;
+        this.name = name;
+        this.color = new RgbaColor(color);
+    }
+
+    public OnlinePlayerMetadata(Client client) {
+        this(client.getId(), client.getName(), client.getColor());
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj instanceof OnlinePlayerMetadata) {
+            OnlinePlayerMetadata other = (OnlinePlayerMetadata) obj;
+            return id.equals(other.id) && name.equals(other.name) && color.equals(other.color);
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, name, color);
+    }
+
+    private final class RgbaColor {
+        private int r;
+        private int g;
+        private int b;
+        private int a;
+
+        private RgbaColor(Color source) {
+            r = source.r();
+            g = source.g();
+            b = source.b();
+            a = source.a();
+        }
+
+        @Override
+        public boolean equals(Object obj) {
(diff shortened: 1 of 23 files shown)


src/main/java/org/terasology/web/resources/OnlinePlayerMetadata.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, name, color);
+    }
+
+    private final class RgbaColor {
```
msteiger, line 53: What is this class needed for? Just for serialization from JSON? Would it work with the NUI class Color if the Color type adapter was registered in the GSON instance?

src/main/java/org/terasology/web/resources/OnlinePlayersResource.java
```diff
@@ -52,11 +53,9 @@ public void onDisconnected(DisconnectedEvent event, EntityRef entityRef) {
...
 
     @Override
-    public List<String> read(Client requestingClient) {
-        List<String> result = new ArrayList<>();
-        for (Client client: networkSystem.getPlayers()) {
-            result.add(client.getName());
-        }
-        return result;
+    public List<OnlinePlayerMetadata> read(Client requestingClient) {
+        return StreamSupport.stream(networkSystem.getPlayers().spliterator(), true)
```
msteiger, line 57: Can you not just call `getPlayers().stream()` here instead? Sorry, I don't have the IDE here, atm.

msteiger (approved):
Very nice, thank you.

src/main/java/org/terasology/web/resources/OnlinePlayersResource.java
gianluca-nitti: Just checked, `getPlayers()` returns an `Iterable<Client>` ([source](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/network/NetworkSystem.java#L50-L50)) which hasn't the `.stream()` method. So the only way I found to get a stream was using StreamSupport (I read it [here](https://stackoverflow.com/a/23936723)).

An alternative could be a plain old foreach which iterates over the `getPlayers()` results and adds the metadata entries to another collection, than return said collection.

src/main/java/org/terasology/web/resources/OnlinePlayerMetadata.java
gianluca-nitti: Yes, the purpose it's serialization. I didn't use the NUI class because the internal representation (which would be serialized) is a single int field storing the 4 channel (r, g, b, alpha) in the 4 bytes of the int, and when it's needed they are extracted with shift operators. So I'd have to copy this logic (the shifts) to the client and when I wrote it I just thought it was easier to make another class with a separate field for each channel.

If you prefer this class to be removed, I could either
- serialize the NUI Color class (or just the representation value, since it's a single number), and separate the channels on the client side, or
- move the fields to the outer `OnlinePlayerMetadata` class, which would then have the fields `String id`, `String name`, `int colorR`, `int colorG`, `int colorB`, `int colorA`

msteiger:
Thanks for the explanation! Merging ..

PR #9, opened 2017-08-25
REST methods

gianluca-nitti:
This branch contains a large rewrite of the protocol-agnostic (HTTP and WebSocket) resource management class structure.

Previously, there was distinction between *writable* and *readable* resource classes, and writable resources used to accept a single method  only (mapped to POST for the HTTP interface).
This was leading to:
- the use unnecessarily complex data structures in the request payloads when accessing resources such as Games or ServerAdmins which have a CRUD nature (games is a collection of savegames that you can read information about, create, and remove; ServerAdmins is a set of client ID strings, so you can read the list and add or remove items; and so on);
- poor usage of the HTTP protocol capabilities (only GET and POST methods);
- an API not following any well-known standard like REST and difficult to document.

The new system is based on REST principles which map more naturally to HTTP without losing the real-time advantage of WebSocket. When using WebSocket, methods are specified as strings and paths as JSON arrays; more information is available in the docs.md document in the root of this branch.

The API specification has also been described in the Swagger/OpenAPI format, in the swagger.json file. It's possible to browse the specification online using this link: http://petstore.swagger.io/?url=https://raw.githubusercontent.com/gianluca-nitti/FacadeServer/rest-methods/src/main/resources/web/swagger.json


README.md
```diff
@@ -6,6 +6,28 @@ Administration comprises starting/stopping the host, configuring modules and the
...
+    * `git checkout develop`
+    * `cd ../..`
+* Run `./gradlew jar`
+* To start a server using FacadeServer, execute `./gradlew facades:Server:run`.
+
+### API documentation
+
+The web server exposes access to the various resources both with an HTTP REST API and over WebSocket.
+More information is available in the docs.md file at the root of this repository.
+Also, here is a [direct link](http://petstore.swagger.io/?url=https://raw.githubusercontent.com/gianluca-nitti/FacadeServer/rest-methods/swagger.json#/) to view online the Swagger/OpenAPI specification of the HTTP API.
```
msteiger: I think that this should be http://editor.swagger.io not "petstore"

src/main/java/org/terasology/web/io/JsonSession.java
```diff
@@ -101,63 +105,60 @@ public static void disconnectAllClients() {
...
     public boolean isAuthenticated() {
         return client != null && !client.isAnonymous();
     }
 
     public ActionResult initAuthentication() {
-        /*if (isAuthenticated()) {
-            return new ActionResult(ActionResult.Status.UNAUTHORIZED, "Already authenticated");
-        }*/
+        if (isAuthenticated()) {
+            return new ActionResult(ActionResult.Status.FORBIDDEN, "Already authenticated");
```
msteiger, line 152: This would change the logic and is thus a bit unrelated to this PR: would it make sense to return `ACTION_NOT_ALLOWED` or even `OK` if somebody tries to auth. again?

msteiger (approved):
Very nice, thank you.

README.md
gianluca-nitti: Sorry for the very late response, I didn't manage to answer your comments before this pull request was merged. Anyway, if necessary I may add changes related to this to the new PR #10.

I used this link because it directs to an interface meant for *viewing* the API doc, which in my understanding is the goal of this link; the editor page shows a split-screen interface which is the same of the petstore on one side and a text editor for the specification file on the other one. Obviously this wouldn't allow people to actually edit the specification file in the repository because it's just fetched through an HTTP GET from `raw.githubusercontent.com`, but anyway I didn't find it ideal for someone which simply needs to view the documentation. This is just my opinion though, if you prefer the editor version I'll switch the link to that one.

Comparison: [Petstore link (currently in the readme file)](http://petstore.swagger.io/?url=https://raw.githubusercontent.com/MovingBlocks/FacadeServer/develop/src/main/resources/web/swagger.json#/) - [Editor link](http://editor.swagger.io/?url=https://raw.githubusercontent.com/MovingBlocks/FacadeServer/develop/src/main/resources/web/swagger.json#/)

src/main/java/org/terasology/web/io/JsonSession.java
gianluca-nitti: Yes I agree that a lot of the possible status codes make sense and I'm not really sure what's the best option... `ACTION_NOT_ALLOWED` maps to [405](https://httpstatuses.com/405) in the HTTP version which - in my understanding - would mean you can't POST /auth, however I don't know - is it meant only for situations where the server *never* supports that method for the requested resource, or it's OK also if it doesn't support it *now*? I suspect there isn't a correct answer because the authentication isn't really stateless (the state is the fact that you are anonymous or authenticated). Also, there is nothing an authenticated client can do with `/auth` (which supports GET and POST for unauthenticated clients, respectively to retrieve the server handshake hello message and to perform the authentication). For `OK` I'm not sure, because, I think (but please correct me if I'm wrong) in the GET part (i.e. an authenticated client trying to redo the authentication handshake) the server should still send a response content.

PR #12, opened 2018-04-03
Add secure websocket support

### LINKED ISSUE #11: Add secure WebSocket support and probably figure out some automated certificate approach
See https://forum.terasology.org/threads/facadeserver-headless-server-with-web-interface.1906/#post-15852 for some context. Might just be some Jetty option? @gianluca-nitti might know more / could figure it out

Ultimately this should allow us to https all the things - communication between this facade, its [front-end](https://github.com/MovingBlocks/FacadeServer-frontend) running on http://utility.terasology.org:8000, the [identity service](https://github.com/MovingBlocks/terasology-key-server) also on http://utility.terasology.org (the default http port), and the [meta server](https://github.com/MovingBlocks/meta-server) on http://meta.terasology.org (for modules)

For certificates likely auto-generate a self-signed cert for this facade and make it easy for the user to replace it with a legit one, like we could do for the instance hosted on utility.terasology.org

Inei1:
### Contains

Fix #11.

So far, I have looked on the internet to see how to add ssl support to jetty. The [Jetty website](http://www.eclipse.org/jetty/documentation/9.4.x/configuring-ssl.html) gives a good starting point. I created a keystore with the keytool like this:
keystoreimg (image is outdated, I made a different keystore with a few changes) as a way to test the server. Jetty supports SSL, so it isn't really that difficult to implement it.

For automatic certificate generation, it looks like we need a keystore. The Java [KeyStore class](https://docs.oracle.com/javase/8/docs/api/java/security/KeyStore.html) provides a way to load a keystore, although to automatically generate what we need for a self-signed certificate we need to do something else. At the moment the certificate I provided seems good enough.

### How to Test

You can test this by trying to login to wss://localhost:8443/ws in the [Facade Server Frontend](http://utility.terasology.org:8000/). Chrome appears to accept connections to the self-signed certificate on localhost, but for other browsers you will need to make an exception. You can also go to https://localhost:8443 in the browser and access the REST APIs over https, such as https://localhost:8443/api/resources/config/MOTD.

### Outstanding Before Merging
- [x] Generate self-signed certificate
- [x] Integrate new code into program better (allow specifying wss port, move some code into other parts if applicable)
- [x] Add a way for server owners to add their own certificate to the server


src/main/java/org/terasology/web/ServerMain.java
```diff
@@ -166,6 +175,26 @@ private static Server createServer(int port, Object... annotatedObjects) throws
...
+        } catch (FileNotFoundException e) {
+            logger.error("Keystore file not found");
+            e.printStackTrace();
+        }
+
+        SslContextFactory sslContextFactory = new SslContextFactory();
+        sslContextFactory.setKeyStorePath(keyStorePath.toString());
+        sslContextFactory.setKeyStorePassword("ServerKeyPassword");
+
+        ServerConnector wssConnector = new ServerConnector(server, new SslConnectionFactory(sslContextFactory,
```
gianluca-nitti: In my understanding this both applies to WebSocket and HTTP, is it correct? If yes I think I would have named it something like `sslConnector` (but it's not a big deal anyway)

src/main/java/org/terasology/web/ServerMain.java
```diff
@@ -166,6 +175,26 @@ private static Server createServer(int port, Object... annotatedObjects) throws
...
+                HttpVersion.HTTP_1_1.asString()), new HttpConnectionFactory());
+
+        wssConnector.setPort(8443);
```
gianluca-nitti: As you mention in the PR message maybe we should make it configurable. I'd just use another env variable like it's done on lines 82-87 for the plain http one (maybe rename `PORT` to `HTTP_PORT` and add an `HTTPS_PORT`

gianluca-nitti (commented):
Looks good to me.

Tested and it worked, however I had to open https://localhost:8443 in the browser and get past the warning before getting it to work with FacadeServer-frontend pointed to wss://localhost:8443/ws in the same browser.

But probably there is not much we can do for this, besides explaining it in documentation. This way if someone hosts a "big" server they could provide their own trusted certificate (from a CA/LetsEncrypt) instead of the self-signed one.

gianluca-nitti:
Also, I checked the 1st checkbox because in my opinion what you did - providing a self-signed certificate - is already good enough, I don't really see any reason to generate one at runtime since it would still be a self-signed one and show warnings in browsers (but please correct me if you disagree).

What is probably more useful is explaining somewhere in the docs how to replace the built-in certificate with a custom one from a CA/LetsEncrypt (I think this is the same thing you meant with your 3rd checkbox), for example by providing some link to documentation about converting other common certificate formats (such as PEM files) to Java Keystore.

Inei1:
I made the appropriate changes. This should be ready for merging if there isn't anything else you can think of.

I created a new keystore file that is more appropriate for end-users. The alias is now serverfacade and the CN is now utility.terasology.org. Apparently the CN is supposed to be the same as the web address, although there will be complaints anyway because the certificate is self-signed.

If users are going to be able to add their own certificate, then they should be able to change the password. I added that as an environment variable option as well.

gianluca-nitti:
Okay thank you for the work. Merged.

PR #13, opened 2018-05-21
Add system status information to REST API

Inei1:
This PR adds relevant information on the system (CPU usage percentage, memory available/total/used/used percentage, system uptime) to the REST API. To view it, make a GET request to api/resources/system with the web browser or another tool, and notice how the data you get in JSON format describes system information, which you can confirm with task manager or another resource manager.

Oshi, the library used here, uses JNA. Luckily, the main Terasology engine already uses JNA in a few places, so we don't need to add any natives or anything like that. The Terasology engine uses a slightly outdated version of JNA, so we have two options:
1. Use an older version of Oshi (3.4.0 instead of 3.5.0) I don't think this will do any harm. This PR was made with this method, and it already works.
2. Update the version of JNA in the engine. I think this will be more work than it is worth for now.

frontend PR: https://github.com/MovingBlocks/FacadeServer-frontend/pull/3

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,8 @@ dependencies {
 
     compile group: 'javax.servlet', name: 'javax.servlet-api', version: '3.1.0'
 
+    compile group: 'com.github.oshi', name: 'oshi-core', version: '3.4.0'
+
     compile group: 'org.glassfish.jersey.containers', name: 'jersey-container-jetty-servlet', version: jerseyVersion
     compile group: 'org.glassfish.jersey.ext', name: 'jersey-mvc-freemarker', version: jerseyVersion
 
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,7 @@ dependencies {
 
     compile group: 'javax.servlet', name: 'javax.servlet-api', version: '3.1.0'
 
+    //TODO: update this to newer version if the engine updates its version of JNA
     compile group: 'com.github.oshi', name: 'oshi-core', version: '3.4.0'
 
     compile group: 'org.glassfish.jersey.containers', name: 'jersey-container-jetty-servlet', version: jerseyVersion
--- a/src/main/java/org/terasology/web/ServerMain.java
+++ b/src/main/java/org/terasology/web/ServerMain.java
@@ -94,7 +94,7 @@ public final class ServerMain {
         }
 
         String keystorePassword = System.getenv("KEYSTORE_PASSWORD");
-        if (keystorePassword == null){
+        if (keystorePassword == null) {
             keystorePassword = "ServerKeyPassword";
             logger.warn("Environment variable 'KEYSTORE_PASSWORD' not defined - using default {}", keystorePassword);
         }
--- a/src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
+++ b/src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
@@ -18,7 +18,8 @@ package org.terasology.web.resources.systemStatus;
 import oshi.SystemInfo;
 import oshi.hardware.HardwareAbstractionLayer;
 
-import java.lang.management.*;
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
 
 
 /**
@@ -30,21 +31,17 @@ public final class SystemMetadata {
     private static SystemMetadata instance;
     private static HardwareAbstractionLayer hardware;
     private static RuntimeMXBean runtimeBean;
-    private static MemoryMXBean memoryBean;
 
     private double cpuUsage;
-    private double memoryUsagePercentage;
     private long memoryUsed;
-    private long memoryTotal;
-    private long memoryAvailable;
+    private long memoryMax;
     private long serverUptime;
     private long jvmMemoryUsed;
-    private long jvmMemoryTotal;
+    private long jvmMemoryMax;
 
     private SystemMetadata() {
         hardware = new SystemInfo().getHardware();
         runtimeBean = ManagementFactory.getRuntimeMXBean();
-        memoryBean = ManagementFactory.getMemoryMXBean();
     }
 
     public static SystemMetadata getInstance() {
@@ -58,37 +55,36 @@ public final class SystemMetadata {
     private void refresh() {
         //Runtime
         cpuUsage = hardware.getProcessor().getSystemCpuLoad() * 100;
-        memoryAvailable = hardware.getMemory().getAvailable();
-        memoryTotal = hardware.getMemory().getTotal();
-        memoryUsed = memoryTotal - memoryAvailable;
-        memoryUsagePercentage = ((double) memoryUsed / memoryTotal) * 100;
+        memoryMax = hardware.getMemory().getTotal();
+        memoryUsed = memoryMax - hardware.getMemory().getAvailable();
         // system uptime in milliseconds
         serverUptime = runtimeBean.getUptime();
         jvmMemoryUsed = Runtime.getRuntime().totalMemory();
-        jvmMemoryTotal = Runtime.getRuntime().maxMemory();
+        jvmMemoryMax = Runtime.getRuntime().maxMemory();
     }
 
     public double getCpuUsage() {
         return cpuUsage;
     }
 
-    public double getMemoryUsagePercentage() {
-        return memoryUsagePercentage;
-    }
-
     public long getMemoryUsed() {
         return memoryUsed;
     }
 
-    public long getMemoryTotal() {
-        return memoryTotal;
-    }
-
-    public long getMemoryAvailable() {
-        return memoryAvailable;
+    public long getMemoryMax() {
+        return memoryMax;
     }
 
     public long getServerUptime() {
         return serverUptime;
     }
+
+    public long getJvmMemoryUsed() {
+        return jvmMemoryUsed;
+    }
+
+    public long getJvmMemoryMax() {
+        return jvmMemoryMax;
+    }
+
 }
--- a/src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
+++ b/src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.web.resources.systemStatus;
+
+import oshi.SystemInfo;
+
+public final class SystemMetadata {
+
+    private double cpuUsage;
+    private double memoryUsagePercentage;
+    private long memoryUsed;
+    private long memoryTotal;
+    private long memoryAvailable;
+    private long systemUptime;
+
+    public static SystemMetadata build() {
+        SystemMetadata systemMetadata = new SystemMetadata();
+        systemMetadata.getStatus();
+        return systemMetadata;
+    }
+
+    private void getStatus() {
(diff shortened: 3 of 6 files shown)


src/main/java/org/terasology/web/ServerMain.java
```diff
@@ -94,7 +94,7 @@ public static void main(String[] args) throws Exception {
         }
 
         String keystorePassword = System.getenv("KEYSTORE_PASSWORD");
-        if (keystorePassword == null){
+        if (keystorePassword == null) {
```
Inei1, line 97: Just fixing a problem with line spacing in my previous PR, #12.

build.gradle
```diff
@@ -33,6 +33,8 @@ dependencies {
 
     compile group: 'javax.servlet', name: 'javax.servlet-api', version: '3.1.0'
 
+    compile group: 'com.github.oshi', name: 'oshi-core', version: '3.4.0'
```
Cervator: Mind adding a note here about how this version was picked since engine has older JNA? Maybe a TODO for updating after the engine JNA updates sometime

src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2017 MovingBlocks
```
Cervator, line 2: Trivial: 2018 (and another place or two)

src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
```diff
@@ -0,0 +1,69 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.web.resources.systemStatus;
+
+import oshi.SystemInfo;
+
+public final class SystemMetadata {
```
Cervator: Some quick javadoc would be awesome, at the class level at least, even for fairly obvious classes :-)

Cervator (approved):
Good stuff! Looks good to me, a few tiny comments added :-)

src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
```diff
@@ -0,0 +1,69 @@
...
+        cpuUsage = systemInfo.getHardware().getProcessor().getSystemCpuLoad() * 100;
+        memoryAvailable = systemInfo.getHardware().getMemory().getAvailable();
+        memoryTotal = systemInfo.getHardware().getMemory().getTotal();
+        memoryUsed = memoryTotal - memoryAvailable;
+        memoryUsagePercentage = ((double) memoryUsed / memoryTotal) * 100;
+        // system uptime in seconds
+        systemUptime = systemInfo.getHardware().getProcessor().getSystemUptime();
+    }
+
+    public double getCpuUsage() {
```
gianluca-nitti: In my understanding, the fields of this class are only accessed through serialization (when the class is serialized to JSON which is sent to the client) and the getters aren't used. Is there any planned use for them? I'd remove them unless there is a reason

src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
```diff
@@ -0,0 +1,69 @@
...
+public final class SystemMetadata {
+
+    private double cpuUsage;
+    private double memoryUsagePercentage;
+    private long memoryUsed;
+    private long memoryTotal;
+    private long memoryAvailable;
+    private long systemUptime;
+
+    public static SystemMetadata build() {
```
gianluca-nitti: Instead of building a new instance on every request, I would make the class a [Singleton](https://en.wikipedia.org/wiki/Singleton_pattern) and then refresh the data when it is requested, e.g.
```
private static final SystemMetadata INSTANCE = new SystemMetadata();

public static SystemMetadata getInstance() { //this is your build()
    INSTANCE.getStatus(); //or INSTANCE.refresh() (see below)
    return INSTANCE;
}
```
This avoids the overhead of object creation (`new SystemMetadata();`) on every request (probably very small and this is a micro-optimization but still makes sense in my opinion).

src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
```diff
@@ -0,0 +1,69 @@
...
+        SystemMetadata systemMetadata = new SystemMetadata();
+        systemMetadata.getStatus();
+        return systemMetadata;
+    }
+
+    private void getStatus() {
```
gianluca-nitti: I'd call it something like `refresh()` - from a `getSomething()` method I'd expect an actual getter which
* doesn't modify the state of the class
* returns something, not `void`

src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
```diff
@@ -0,0 +1,69 @@
...
+        SystemInfo systemInfo = new SystemInfo();
```
gianluca-nitti: Oh and to go on with the micro-optimizations, this might be stored as an instance field, i.e. `private SystemInfo systemInfo = new SystemInfo();` and only initialized once together with this class (in the declaration, i.e. no other `new`s besides the one in the snippet I just wrote), **but** the implementation of `SystemInfo` may either
1. read the data from the system when the instance is constructed, store it in its internal fields and then always return the same data when the getters are called, or
2. obtain the data every time the getters are called
Obviously, my suggestion would only make sense if it works as in option 2. Having a look at oshi's docs/source and/or making some empirical observation whether we always get the same data could be helpful to understand if it works like 1 or 2 and thus if it makes sense to apply this change.

However if it's too much work it may not be worth it. These are mostly random thoughts and is by no means "requested changes"

And finally - in the local variable or instance field (`systemInfo` but name should we changed as well if the type is changed) you may store the result of `getHardware()` since it's its only member we need.
My reasoning is that since oshi is an external library we don't really know what `getHardware()` actually does - it may make expensive calculations (I don't think it's the case but who knows) so if we can reduce the amount of calls we do to external code why not? (of course there may be counterarguments, happy to hear them if anyone has any)

gianluca-nitti (commented):
Thank you @Inei, good work!

Left some comments with some suggestions, and a general one here: if I understand correctly, the provided stats are all system wide. If it's feasible and not too time-consuming, it may be useful to also send some stats about the running JVM so that one can see which system resources *the game* is actually using? For example data about the [heap size/usage](https://stackoverflow.com/questions/2015463/how-to-view-the-current-heap-size-that-an-application-is-using) (suggesting this since if I remember correctly issues with small max heap size is what usually causes out-of-memory errors in java programs rather than amount of physical memory, so it may be useful to show this kind of info as well)

mkienenb:
Just skimming the PR and the slack channel, but I will note that https://github.com/patric-r/jvmtop provides some pretty interesting stats and I think it's all native java code.

src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
Inei1: I plan on writing a unit test at some point that will need the getters.

src/main/java/org/terasology/web/resources/systemStatus/SystemMetadata.java
```diff
@@ -16,54 +16,75 @@
...
-        SystemMetadata systemMetadata = new SystemMetadata();
-        systemMetadata.getStatus();
-        return systemMetadata;
+    private long memoryMax;
+    private long serverUptime;
+    private long jvmMemoryUsed;
+    private long jvmMemoryMax;
+
+    private SystemMetadata() {
+        hardware = new SystemInfo().getHardware();
```
gianluca-nitti: A small note: it doesn't look right to me to initialize static fields inside a constructor. The static fields belong to the **class**, while the constructor builds an **object** / **instance** of the class.
In this case we have a singleton so there will always be only one instance, so there are no practical consequences, but I still think that from a logical point of view it would be better to
* either keep the fields static and initialize them when the class is loaded (not when the instance is created) - this can be done either in the declaration (`private static HardwareAbstractionLayer hardware = new SystemInfo().getHardware();`) or in a static initialization block (more info [here](https://docs.oracle.com/javase/tutorial/java/javaOO/initial.html), 1st option should be enough since they are just simple assignments), or
* make the two fields instance member (e.g. not static). This would mean "every" instance of `SystemMetadata` has its own `hardware` and `runtimeBean` (but in reality, there is only one instance, so it shouldn't make any difference in practice), and thus it's OK to initialize them in the constructor. I think the only necessary change would then be marking the fields as [transient](https://stackoverflow.com/a/5889590) so that they won't be serialized by gson.

src/main/java/org/terasology/web/resources/systemStatus/SystemResource.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.terasology.web.resources.base.AbstractSimpleResource;
+import org.terasology.web.resources.base.ClientSecurityRequirements;
+import org.terasology.web.resources.base.ResourceAccessException;
+import org.terasology.web.resources.base.ResourceMethod;
+import org.terasology.web.resources.base.ResourcePath;
+
+import static org.terasology.web.resources.base.ResourceMethodFactory.createParameterlessMethod;
+
+public class SystemResource extends AbstractSimpleResource {
```
gianluca-nitti: I tested the whole server+frontend setup and it mostly works as expected, except one detail - the frontend shows the values received at the moment of the connection, and they are not updated. I noticed you added `system` as one of the resources the home tab is subscribed to, so the frontend should be ready to receive updates without asking for them; now we need to have the server actually send these updates.

I'd suggest you take a look (if you haven't already), for example, at `OnlinePlayersResource`. As you can see, when the value changes and you want to send the update to all the connected clients, you just have to call `notifyChangedForAllClients()` (which is a method of the `AbstractSimpleResource` class).

The only difference is that `OnlinePlayersResource` knows when there is an update to send: when it receives an event from the engine which says a player has joined or disconnected. On the other hand, `SystemResource` doesn't really know when to send updated stats, since they are changing all the time. I'd suggest to send them periodically, for example every 5 seconds (or maybe we can make it configurable).

To implement this you may use a thread which runs in the background and runs an infinite loop in which it sleeps for a certain amount of seconds, then calls `notifyChangedForAllClients()`. But there are probably better ways, for example there is a class called `ScheduledExecutorService` which if I understand correctly should be well-suited for this purpose. Googling for "java schedule periodic task" leads to various interesting results ([here is one](https://stackoverflow.com/questions/7814089/how-to-schedule-a-periodic-task-in-java)).

The whole resource handling system **should** be thread safe - I think calls to `onConnected` and `onDisconnected` in `OnlinePlayersResource` also come from a separate thread (not the same one the web server is running in) - however in case you run into problems I think we can merge the pull request at this point and fix the push notifications later.

gianluca-nitti (commented):
I think it's almost ready to merge.

Left a couple of comments, one is mostly a code style suggestion (about static vs instance fields initialization), the other is a hint about how to periodically send updated values to the client.

src/main/java/org/terasology/web/resources/systemStatus/SystemResource.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final ScheduledExecutorService refreshSystemInfoService = Executors.newScheduledThreadPool(1);
+
+    public void startSystemInfoRefreshService() {
+        refreshSystemInfoService.scheduleAtFixedRate(this::refreshSystemInfoForClient, 1000, 1000, TimeUnit.MILLISECONDS);
+    }
+
+    @ReceiveEvent
```
gianluca-nitti, line 39: Just one more note: the `@ReceiveEvent` annotation [is defined in the engine](https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/entitySystem/event/ReceiveEvent.java), and is used to mark methods inside Component System classes which can be used to handle events.
If you notice, in `OnlinePlayersResource`, `onConnected` and `onDisconnected` are never called directly (except in unit tests) - it's the engine which discovers them through reflection because the class implements the `ComponentSystem` interface and is annotated with `@RegisterSystem`, and calls them when appropriate. If you are interested in the topic, you may find [this page in the Terasology's wiki](https://github.com/MovingBlocks/Terasology/wiki/Events-and-Systems) useful.

It is not useful outside of `ComponentSystem`s.

gianluca-nitti (approved):
Okay, good to merge.
Left one more comment, but I'll fix that together with other small refinements like the years in the license texts.

PR #14, opened 2018-05-23
Gradle tweaks to allow standalone/embedded workspace builds/usage

Cervator:
Improves but doesn't fully clean up the current Gradle setup. Tested in the Nanoware fork, Jenkins, and Artifactory.

* http://jenkins.terasology.org/view/Nanoware/job/NanoFacadeServer/6/console
* http://jenkins.terasology.org/view/Nanoware/job/NanoFacadeServer/lastSuccessfulBuild/artifact/build/libs/
* http://artifactory.terasology.org/artifactory/webapp/#/artifacts/browse/tree/General/nanoware-snapshot-local/org/terasology/web/NanoFacadeServer/1.0.0-SNAPSHOT
* http://jenkins.terasology.org/view/Ext/job/FacadeServer (when merged)

Note the location of stuff will change as this is merged to where the regular job will build it instead of the experimental jobs.

Goes with https://github.com/MovingBlocks/Terasology/pull/3371 to fix a pathing issue when the engine is used as a retrieved jar dependency.

Pinging @Inei1 as I can't find via Reviewer dropdown

The regular Jenkins job should be up to date and able to run properly after this is merged.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,18 @@
+// Dependencies needed for what our Gradle scripts themselves use. It cannot be included via an external Gradle file :-(
+buildscript {
+    repositories {
+        // External libs - jcenter is Bintray and is supposed to be a superset of Maven Central, but do both just in case
+        jcenter()
+        mavenCentral()
+    }
+
+    dependencies {
+        // Artifactory plugin
+        classpath(group: 'org.jfrog.buildinfo', name: 'build-info-extractor-gradle', version: '4.0.0')
+
+    }
+}
+
 plugins {
     id "org.standardout.versioneye" version "1.3.0"
 }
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -96,9 +119,15 @@ task setupServerModules(type: Sync) << {
...
-run.workingDir = rootDir
-run.args = ["-homedir=terasology-server"]
\ No newline at end of file
+
+// Support both standalone workspace (expect certain things provided) and embedding in a Terasology workspace (use task dependencies)
+if (project.name != project(':').name) {
+    run.dependsOn rootProject.extractNatives
+    run.dependsOn rootProject.moduleClasses
+} else {
+    // For a standalone workspace natives should be provided
```
gianluca-nitti, line 132: So if I understand correctly this is the only part which is missing to have tests work in a standalone workspace, correct?
Would it work if we just copy the relevant parts (i.e. `dependencies {natives group: 'org.lwjgl.lwjgl', ...}`, `task extractNatives {...}`) from the engine's `build.gradle`, then add `dependsOn extractNatives` here?

As a (maybe better but probably more time consuming) alternative - could the code which takes care of natives be moved from the engine's `build.gradle` to somewhere accessible from the various repositories, like the `config` dir which Jenkins copies into the workspace before starting the build? Maybe it could be useful for testing modules as well? (I always worked on the engine and on this facade, but never on a module, so I don't know if they have unit tests which run during the Jenkins build)

EDIT: I read description of MovingBlocks/Terasology#3371 after writing this comment, now I realize just copying the natives directory is the simple option since it's already available on the build server

gianluca-nitti (approved):
Thank you @Cervator - I think this is quite a big step, with this patch I was able to finally build a jar (with `./gradlew jar`) in a standalone environment, i.e. just by cloning this repo (without the engine repo) and copying the config directory. I also tested the usual setup (this repo in the `facades/Server` dir of the engine repo) and it still works correctly.

I think it would be nice if we could run the unit tests too. In my understanding the obstacle are the missing natives - if I run `./gradlew check` the various static checking tools (checkstyle, PMD, FindBugs) work correctly, but when it gets to the tests, as expected it complains about missing natives because tests need to run the engine which in turn needs the natives.

However if it's a substantial amount of work maybe we can merge this now so that we can build jars on Jenkins, which could then be downloaded by the launcher.

PR #15, opened 2018-05-27
Expand Javadoc for the backend

Inei1:
This is a work in progress, adding some Javadoc to the classes that I think could have more.


src/main/java/org/terasology/web/authentication/HandshakeHello.java
```diff
@@ -22,6 +22,9 @@
 
 import java.nio.ByteBuffer;
 
+/**
+ * Performs the handshake hello used to communicate between the client and the server.
```
gianluca-nitti: I wouldn't say this class actually *performs* anything (the `validate()` method just does some null checks to avoid NPEs to be thrown later) - it's mostly for storing an handshake hello message. The actual handshake implementation is in [`AuthenticationHandshakeHandlerImpl`](https://github.com/Inei1/FacadeServer/blob/41810711438c8785ed92653f23fc020eb7f61438/src/main/java/org/terasology/web/authentication/AuthenticationHandshakeHandlerImpl.java).
So here I'd say
> **Represents an** handshake hello **message** used to communicate between the client and the server.

src/main/java/org/terasology/web/client/AuthenticatedHeadlessClient.java
```diff
@@ -20,6 +20,12 @@
 import org.terasology.network.internal.AbstractClient;
 import org.terasology.rendering.nui.Color;
 
+/**
+ * This class is used to make an authenticated client, which is able to connect to the
+ * server. Note that authenticated does not necessarily mean that the client can perform admin
```
gianluca-nitti, line 25: Good - I think it's important to point this out

src/main/java/org/terasology/web/client/HeadlessClient.java
```diff
@@ -24,6 +24,9 @@
 import org.terasology.web.serverAdminManagement.ServerAdminsManager;
 import org.terasology.world.chunks.Chunk;
 
+/**
+ * Interface for clients attempting to access resources or connect to the game.
```
gianluca-nitti: I think the "or connect to the game" can be a bit misleading. This interface extends the engine's `Client` interface to represent an headless client (which is a client connected through the web interface, without a normal game client) - maybe the comment can be rewritten to point this out

src/main/java/org/terasology/web/resources/DefaultComponentSystem.java
```diff
@@ -17,6 +17,9 @@
 
 import org.terasology.entitySystem.systems.ComponentSystem;
 
+/**
+ * Interface for component systems, same as {@link org.terasology.entitySystem.systems.ComponentSystem} without the autosave methods.
```
gianluca-nitti: I think it's more clear in this way:
```
Extension of {@link org.terasology.entitySystem.systems.ComponentSystem} which provides default
do-nothing implementations of the various preBegin(), postSave(), etc methods, so that,
for convenience, implementor classes only need to override the necessary methods
```

src/main/java/org/terasology/web/resources/base/ResourceMethodName.java
```diff
@@ -15,6 +15,9 @@
  */
 package org.terasology.web.resources.base;
 
+/**
+ * This enum contains the five HTTP requests used by the server.
```
gianluca-nitti: As said above they are HTTP request **methods**.

src/main/java/org/terasology/web/resources/base/ResourceMethod.java
```diff
@@ -18,11 +18,32 @@
...
+     * @return whether the client has the permissions required.
+     */
     boolean clientIsAllowed(ClientSecurityInfo securityInfo);
 
+    /**
+     * Perform the HTTP request for the client.
+     * @param data the data to send to the server (if any).
+     * @param client the client that requested the perform.
+     * @return the data requested by the server.
+     * @throws ResourceAccessException the specified resource cannot be retrieved.
```
gianluca-nitti: I'm sorry if I'm being a bit pedantic, but I think it's better to use accurate wording:
* (1st line) `data` is the data **sent** to the server by the client - saying *to send* may be misleading because we are on the server here, we already received this data
* (2nd line) I could be wrong but isn't *perform* a verb? I'd say `the client that performed the request` 
* (3rd line) The client has made a request and the server is answering it with the data returned by this method. So I'd say either `the data requested by the client` or `the answer to the client's request`

src/main/java/org/terasology/web/resources/base/ResourceMethod.java
```diff
@@ -18,11 +18,32 @@
 import org.terasology.network.Client;
 import org.terasology.web.client.ClientSecurityInfo;
 
+/**
+ * Interface used to perform a resource request.
+ * @param <INTYPE> the parameter received from the client.
+ * @param <OUTTYPE> the parameter given to the client.
+ */
```
gianluca-nitti: To make it more descriptive, I think it should be specified that these are the types used for **deserialization** (INTYPE) and **serialization** (OUTTYPE) of the data, respectively, received from and sent to the client.

src/main/java/org/terasology/web/resources/base/ResourceAccessException.java
```diff
@@ -17,6 +17,10 @@
 
 import org.terasology.web.io.ActionResult;
 
+/**
+ * This exception is thrown when attempting to either access a resource which does not exist or
+ * when trying to perform a HTTP request on a resource that is not supported
+ */
```
gianluca-nitti: There are various different situations in which this exception can be thrown. So I'd specify that these are examples. Also, the second line isn't very clear to me - I'd say `when the requested resource doesn't support the specified method`.

While the method names (GET, POST, PUT, etc) are taken from the HTTP standard, they are also used with WebSockets (the request method is specified as part of the RESOURCE_REQUEST message) - they are not HTTP-specific.

src/main/java/org/terasology/web/resources/base/AbstractObservableResource.java
```diff
@@ -17,6 +17,10 @@
 
 import org.terasology.entitySystem.entity.EntityRef;
 
+/**
+ * Basic implementation of the resource interface for resources that are exposed over REST,
+ * taking all methods from the resource observer.
+ */
```
gianluca-nitti: I think a more clear description would be
```
Base class for a Resource (maybe put here @link to the Resource interface) which can notify an
attached observer of events or changes in it's data.
```

src/main/java/org/terasology/web/resources/base/AbstractSimpleResource.java
```diff
@@ -15,8 +15,18 @@
...
 public abstract class AbstractSimpleResource extends AbstractObservableResource {
 
+    /**
+     * Determine which HTTP request method to use based on the methodName.
+     * @param methodName type of http request (GET, PUT, etc.).
+     * @param path path of the URL to the resource (api/resources/console for example).
+     * @return the method used to request the resource.
+     * @throws ResourceAccessException the requested method is not supported by this resource.
+     */
     @Override
```
gianluca-nitti: I think it's not necessary to duplicate the javadoc comment when overriding methods (unless you want to describe the specific implementation provided by this class, in which case you write a comment which is different from the one in the parent class). I think if you leave it empty, when you generate the javadoc it will automatically take the comment from the superclass (see [here](https://stackoverflow.com/a/15906009))

src/main/java/org/terasology/web/resources/base/Resource.java
```diff
@@ -15,11 +15,29 @@
...
 package org.terasology.web.resources.base;
 
+/**
+ * Base interface for the resource implementation.
+ */
 public interface Resource {
 
+    /**
+     * Determine the method used to request a resource.
+     * @param methodName type of http request (GET, PUT, etc.).
```
gianluca-nitti: `GET`, `POST`, etc are HTTP **methods** (see the [RFC here](https://tools.ietf.org/html/rfc7231#page-22)). An HTTP request is the whole message containing the method, the path, and request headers.

gianluca-nitti (commented):
Thank you for all the documentation work.

I left some comments to clarify some concepts, I'm sorry if there are parts of code which weren't very clear in first place - I should have wrote more documentation myself last year.

src/main/java/org/terasology/web/io/gsonUtils/Validable.java
```diff
@@ -15,7 +15,14 @@
  */
 package org.terasology.web.io.gsonUtils;
 
+/**
+ * Interface for validating a client connection certificate.
```
gianluca-nitti: Not necessarily - as you can see, for example `ResourceRequestClientMessage` also implements this interface and has nothing to do with authentication/certificates. This is just a common interface for messages coming from the user (i.e. objects deserialized from json) which may or may not be well-formed and thus can be validated before doing further processing.

So I'd fix this comment and the one for the `validate()` method.

src/main/java/org/terasology/web/resources/base/ParameterlessMethodHandler.java
```diff
@@ -17,6 +17,12 @@
 
 import org.terasology.network.Client;
 
+/**
+ * Method handler for {@link ResourceMethodFactory} which is used for resources where the client needs to get
+ * data from the server after the {@link ResourceMethod} finishes executing. Used in GET methods.
+ * @param <INTYPE> the type of data sent to the server from the client through the {@link ResourceMethod}.
+ * @param <OUTTYPE> the type of data received from server by the {@link ResourceMethod} after executing.
```
gianluca-nitti: The part about `OUTTYPE` is incorrect - we are on the server, and this is the type of the data which will be *sent to the client*.

src/main/java/org/terasology/web/resources/base/ParametrizedMethodHandler.java
```diff
@@ -17,6 +17,12 @@
 
 import org.terasology.network.Client;
 
+/**
+ * Method handler for {@link ResourceMethodFactory} which is used for resources where the client needs to
+ * give extra parameters to the server for execution. Currently unused.
+ * @param <INTYPE> the type of data sent to the server from the client through the {@link ResourceMethod}.
+ * @param <OUTTYPE> the type of data received from server by the {@link ResourceMethod} after executing.
```
gianluca-nitti: Same here. To clarify:
* `INTYPE` is the format of **client to server` data, i.e. the body of a request
* `OUTTYPE` is the format of **server to client** data, i.e. the body of a response

If we use the *sending* and *receiving* terms I think we have to use them from the server point of view since we are in the server's code. So the comment for `INTYPE` is correct, while the ont for `OUTTYPE` is not (I'd replace `received from server` with something like `sent from the server to the client`)

src/main/java/org/terasology/web/resources/base/ResourceMethod.java
```diff
@@ -18,11 +18,32 @@
 import org.terasology.network.Client;
 import org.terasology.web.client.ClientSecurityInfo;
 
+/**
+ * Interface used to perform a resource request.
+ * @param <INTYPE> the parameter received from the client, used for serialization.
```
gianluca-nitti: It's the other way round. You (the server) are receiving `<INTYPE>` data, thus you have to **deserialize** it from JSON (which you receive from an HTTP request body, or a WebSocket message) to a Java object to process it; then, you want to send `<OUTTYPE>` data as a response to the client, so you build it as a Java object and then **serialize** it to JSON (which you then send as an HTTP response or a new WebSocket message).

src/main/java/org/terasology/web/resources/base/ResourceMethodImpl.java
```diff
@@ -18,6 +18,11 @@
 import org.terasology.network.Client;
 import org.terasology.web.client.ClientSecurityInfo;
 
+/**
+ * Implementation of {@link ResourceMethod}.
+ * @param <INTYPE> the type of data sent to the server from the client.
+ * @param <OUTTYPE> the type of data received from server.
```
gianluca-nitti: Another place to fix comment about `OUTTYPE` (see previous comments)

src/main/java/org/terasology/web/resources/base/VoidParameterlessMethodHandler.java
```diff
@@ -17,6 +17,11 @@
 
 import org.terasology.network.Client;
 
+/**
+ * Method handler for {@link ResourceMethodFactory} that is used for resources where the server does not need to
+ * give any data back to the client after the method executes. Used in PUT, POST, PATCH, and DELETE methods.
+ * @param <INTYPE> the type of data sent to the server from the client through the {@link ResourceMethod}.
+ */
```
gianluca-nitti, line 24: Good - as you correctly wrote, in this case no response is sent to the client, and as you can see there is no `OUTTYPE`, so this is an example of the fact that `OUTTYPE` is related to data *sent from the server to the client* and not received by the server

(nothing to change here, this comment is just to further clarify the discussion about `OUTTYPE` from other classes - see the previous comments)

src/main/java/org/terasology/web/resources/games/GamesBackupsResource.java
```diff
@@ -34,6 +34,9 @@
 
 import static org.terasology.web.resources.base.ResourceMethodFactory.createVoidParameterlessMethod;
 
+/**
+ * {@link org.terasology.web.resources.base.ResourceMethod} used for backing up games.
```
gianluca-nitti: Wrong link (probably a typo)  - as you can see from the next line (`public class GamesBackupsResource extends AbstractSimpleResource`), this is a `Resource`, not a `Method`

It's a bit different from the other similar classes (`NewGameMethod`, `PatchGameMethod`, etc) because since there wasn't an appropriate [HTTP method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) for backups (i.e. `POST` is for creating something in a resource so it's good for `NewGame`, `PATCH` is for partially modifying something in a resource so it fits the concept of renaming, etc, but I thought none of the standard methods was appropriate for a backup/duplication), so I moved them to a separate resource in which `POST` means "create a new backup" (various interesting ideas for designing this kind of APIs [here](https://stackoverflow.com/questions/6850187/non-crud-operations-in-a-restful-service))

gianluca-nitti (commented):
Thank you @Inei1 for the documentation work - finally managed to review all of it.

There are some small things to fix, mostly related to the type parameters (`INTYPE` and `OUTTYPE`) - this is probably due to my code not being very clear, I apologize for that.
I now explained everything relevant in the comments, I think it should be easier to understand now but if you have any questions I'll be happy to clarify further.

Also, since we merged some of the other pull requests you made - for example #17 - there ~~are some~~ (EDIT: just noticed that only `ClientSecurityInfo` is affected) is a small merge conflict. A way to fix them is to merge the `develop` branch inside this (using git from the command line), then Git should warn you that there are merge conflicts, which you can solve by editing the files appropriately and then committing. Some useful info can be found [here](https://help.github.com/articles/resolving-a-merge-conflict-using-the-command-line/).

gianluca-nitti:
Everything looks fine to me now - merged!

PR #16, opened 2018-05-28
add a GET method to REST API from console resource to get all commands

Inei1:
Just a simple change to console resource, adding a GET method that gives all commands.

### DIFF
--- a/src/main/java/org/terasology/web/resources/console/ConsoleResource.java
+++ b/src/main/java/org/terasology/web/resources/console/ConsoleResource.java
@@ -15,6 +15,7 @@
  */
 package org.terasology.web.resources.console;
 
+import com.google.common.collect.Collections2;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.event.ReceiveEvent;
 import org.terasology.entitySystem.systems.RegisterSystem;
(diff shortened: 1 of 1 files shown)


src/main/java/org/terasology/web/resources/console/ConsoleResource.java
```diff
@@ -42,6 +46,12 @@ public void onMessage(MessageEvent event, EntityRef entityRef) {
         notifyEvent(entityRef, event.getFormattedMessage());
     }
 
+    @Override
+    protected ResourceMethod<Void, Collection<String>> getGetMethod(ResourcePath path) throws ResourceAccessException {
+        return createParameterlessMethod(path, ClientSecurityRequirements.PUBLIC, Void.class, (data, client) ->
+            Collections2.transform(console.getCommands(), input -> input.getName().toString()));
```
gianluca-nitti: I would apply a filter to only display the commands which run on the server (there are commands such as `join`, `leave`, `playTestSound` which run in the client). Should be very simple since `ConsoleCommand` has a `isRunOnServer` method).
As you may have noticed in other parts of the project this kind of operations are usually done in a functional programming-like way with [streams](https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html), introduced in Java 8. Since we [require Java 8](https://github.com/MovingBlocks/Terasology#playing), we can happily use them instead of Google Collections2 which is an external library.
So I'd suggest something like 
```
console.getCommands().stream()
                    .filter(ConsoleCommand::isRunOnServer)
                    .map(ConsoleCommand::getName)
                    .map(Name::toString)
                    .collect(Collectors.toList()));
```
(if you go this way, some additional imports are required, for example `Name` is from `org.terasology.naming`)

gianluca-nitti (commented):
Thank you, tested and works.

Left a comment about filtering out client-side commands.

PR #17, opened 2018-06-09
Backend for admin permissions

Inei1:
This PR provides the logic for blocking admins from performing certain actions over the REST API.

To test, start up a new server and join the game. There should be a new file called serverAdminPermissions.json. You can also query resources/serverAdminPermissions to get a list of all admin permissions and resources/serverAdmins/{adminId}/permissions for the permissions of a single admin.


src/main/java/org/terasology/web/serverAdminManagement/AdminPermissions.java
```diff
@@ -0,0 +1,114 @@
...
+ */
+package org.terasology.web.serverAdminManagement;
+
+/**
+ * Class intended to be used for serializing admin permission JSON objects over the REST API.
+ */
+public class AdminPermissions {
+
+    private String id;
+    private Permissions permissions;
```
gianluca-nitti: I think an alternative to the custom `Permissions` class could be a `Map<PermissionType, Boolean>`. Or maybe even a `Set<PermissionType>` where you only add the entries you'd set to true in the map (but this may involve a bit more work since it may require to change the frontend; with the map, if gson allows to serialize it as a JSON object of key/value pairs, it may be possible to keep the same format of the serialized data).
I'm suggesting you have a look at this alternatives so that we may reduce the amount of code a bit - for example, right now the fields of `Permissions` and the values of the `PermissionType` enum are the same list, which means that if we want to add a permission we have to change the code in two places. In my opinion it would be nice if we can avoid this kind of duplications when possible.

src/main/java/org/terasology/web/serverAdminManagement/AdminPermissionManager.java
```diff
@@ -0,0 +1,226 @@
...
+                    return true;
+                case POST:
+                    if (path.toString().contains("games")) {
+                        return permissions.getPermission(PermissionType.CREATE_BACKUP_RENAME_GAMES);
+                    } else if (path.toString().contains("serverAdmins")) {
+                        return permissions.getPermission(PermissionType.ADMIN_MANAGEMENT);
+                    }
+                    return true;
+                case PUT:
+                    if (path.toString().compareTo("engineState") == 0) {
```
gianluca-nitti: `.equals()` should work as well

src/main/java/org/terasology/web/serverAdminManagement/AdminPermissionManager.java
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    public static AdminPermissionManager getInstance() {
+        if (instance == null) {
+            instance = new AdminPermissionManager(PathManager.getInstance().getHomePath().resolve("serverAdminPermissions.json"));
+        }
+        return instance;
+    }
+
+    public boolean adminHasPermission(String adminID, ResourcePath path, ResourceMethodName resourceMethodName) {
```
gianluca-nitti: I think a better implementation could be if the resources themselves - or better, each method of each resource - knew which permissions are required to access them. For example, the way I checked for permissions (which, as you already know, were just binary before your code - a client was either an admin or a non-admin, there were no permission levels) made use of the `ClientSecurityRequirements` class. An instance of this class is created for each resource method and passed to the method builders (in `ResourceMethodFactory`, like `createVoidParameterlessMethod` and similar), and encapsulates what are the requirements to execute a certain method (`GET`, `POST`, etc) on a certain resource. Maybe you could modify `ClientSecurityRequirements` to also encapsulate which permission type is required, and have its `clientIsAllowed` method somehow query `AdminPermissionManager` to determine its result?

gianluca-nitti (commented):
Thank you - looks very good.

In the comments I'm suggesting some refactorings which could make the codebase easier to work with.
They are probably not minor (especially the one about reimplementing `adminHasPermission`) so I suggest you evaluate whether they are worth it - in my personal opinion they are, but you may not agree especially considering you are the one actually doing the work :)
So as always if you have counterarguments, happy to hear them

Inei1:
~~I am in the process of refactoring some of the code for this PR, so don't merge this (or the frontend PR) yet. None of the logic should be changed, so you could probably test it.~~

The refactoring is done, this PR is fine to test now. There is a bug in the implementation for the admin permissions for the console, and taking away the permissions after giving them doesn't appear to prevent users from executing console commands on the frontend only.

src/main/java/org/terasology/web/client/ClientSecurityInfo.java
```diff
@@ -18,17 +18,28 @@
 public class ClientSecurityInfo {
 
     private boolean isAuthenticated;
+    private boolean isAdmin;
```
gianluca-nitti: Might be good to put a short commend giving a hint about what's the difference between being an admin and having admin permissions

gianluca-nitti (commented):
Thank you @Inei. Lots of good code overall.

If it's not too much work, and if you agree, here is an idea for a refactoring which in my opinion would make everything more readable and easier to maintain.

If I understand correctly, right now `ClientSecurityInfo` encapsulates information (a boolean) about just one of the new permissions, and an object of `ClientSecurityRequirements` stores if  acertain permission is required but without knowing which one.

I think a better approach would be having `ClientSecurityInfo` storing the list of permissions the user has (e.g. in a `List<PermissionType> ownedPermissions`), and `ClientSecurityRequirements` storing the required permission in a `PermissionType requiredPermission` field. This way, `clientIsAllowed` in `ClientSecurityRequirements` could check if the client has the required permission in it's list (you could add an `boolean hasPermission(PermissionType p)` in `ClientSecurityInfo`, or a getter which returns a read-only view of the list, etc - there are many ways).

I think this could be a good idea since you would encapsulate more logic in `ClientSecurityInfo` and `ClientSecurityRequirements`, and wouldn't need all the changes in the `ResourceMethod` classes (e.g. `getPermissionType()`) and their builders (`createParametrizedMethod`, etc) since everything regarding security and permissions would be handled by the two dedicated classes, which are already passed around when necessary.

src/main/java/org/terasology/web/serverAdminManagement/PermissionType.java
```diff
@@ -0,0 +1,48 @@
...
+    CONSOLE_SERVER_MANAGEMENT,
+    CONSOLE_DEBUG,
+    INSTALL_MODULES,
+    CREATE_BACKUP_RENAME_GAMES,
+    DELETE_GAMES,
+    START_STOP_GAMES,
+    CHANGE_SETTINGS,
+    ADMIN_MANAGEMENT;
+
+    public static Map<PermissionType, String> getConsolePermissionsMap() {
```
gianluca-nitti, line 40: I think using a map is a good idea. An alternative IMHO could be storing the mapped console permission inside a field of the instance of the enum, e.g. having a `private String mappedConsolePermission;` (remember that an enum is just like another class, it just has a fixed number of instances - the enum entries; see for example the [planets example here](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html)), or maybe it could be an `Optional<String>` since some entires have no mapped console permission. Then you would assign it in the constructor, and you could write the entries like `CONSOLE_CHEAT(PermissionManager.CHEAT_PERMISSION), START_STOP_GAMES(Optional.empty())` (or `null` if you don't use `Optional`, or make an overloaded constructor which takes no arguments for the values which have no mapping, etc - various possibilities).

Then you would be able to know the mapped console permission via a, for example, `getMappedConsolePermission()` method.

However I see that you call `getConsolePermissionsMap()` just once and you need the mappings for all of them at once - so I'd say your way is equally good. Just take it as a hint for similar situations if you prefer keeping it this way

src/main/java/org/terasology/web/resources/ResourceManager.java
```diff
@@ -94,9 +97,11 @@ public void initialize(TerasologyEngine gameEngine) {
                         .addSubResource("MOTD", new ServerMotdResource())
                         .build())
                 .addSubResource("serverAdmins", new ServerAdminsResource())
+                .addSubResource("serverAdminPermissions", new AdminPermissionListResource())
                 .addSubResource("system", systemResource)
                 .build();
         systemResource.startSystemInfoRefreshService();
+        InjectionHelper.inject(AdminPermissionManager.getInstance(), context);
```
gianluca-nitti, line 103: `@In`s and `InjectionHelper` are mostly useful when you have to inject the values in a lot of classes, to avoid the verbosity of passing them in each constructor and similar situations. However here there is just one class, so I'd have just put something like a `setNetworkSystem` method and called it here (or passed it to the constructor, but not really feasible in this case since it's a singleton). You should be able to obtain the `NetworkSystem` from `context` with a `context.get(NetworkSystem.class)` - which is basically what `InjectionHelper` does IIRC, it looks at the types of the fields to be injected and queries the context for an instance of those classes

gianluca-nitti (commented):
Here are a few more small comments.

Also I have a question, do you think it would make sense to merge `ServerAdminsManager` and `AdminPermissionManager`, together? In my understanding your new `AdminPermissionManager` is a way more advanced version of my old `ServerAdminsManager`, do you think it would make sense to integrate the parts of `ServerAdminsManager` which are still necessary into `AdminPermissionManager` and then remove `ServerAdminsManager`?

Just wondering if we can avoid splitting code used for similar purposes in multiple classes with similar functionality - what's your opinion on this? Anyway maybe not right now since you probably already have lots of stuff to do, maybe you could do it in buffer/stretch weeks or I can do it myself after your GSoC

src/main/java/org/terasology/web/resources/ResourceManager.java
Inei1: Is it still worth it to do this now that I have added another injected item?

Inei1:
I have done the requested refactoring, and it conveniently fixed one of the more confusing bugs I was dealing with. As for merging the two admin manager classes, I think they could work together. The reason I didn't do it at first is because I wanted to simplify the code. I don't think it would be difficult to merge the two later.

Hold off on merging this for now as I still need to add to swagger.json.

src/main/java/org/terasology/web/resources/ResourceManager.java
gianluca-nitti: Probably not, I think it's fine to keep it this way

src/main/java/org/terasology/web/resources/serverAdmins/AdminPermissionListResource.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.web.resources.serverAdmins;
+
+import javafx.util.Pair;
```
gianluca-nitti: Sorry, I didn't notice before (I assumed `Pair` was a class you made or a class from some of the utility libraries we use, like Guava) - but the sue of this class is problematic.

I was going to test the code (I didn't have the time to do it in the time range between usage of `Pair`s was introduced and today), however compiling failed with a `package javafx.util does not exist`.
So if I remember correctly you are on Windows so it probably works fine there since JavaFX is bundled with the Oracle JDK (or even JRE, I don't remember) installers. However, OpenJDK - which is the Java environment most people uses on Linux (I don't know how the situation is for macOS) - doesn't bundle it (due to licensing issues IIRC) - it's available as a [separate package](http://openjdk.java.net/projects/openjfx/).

We *could* add JavaFX as a requirement (and instruct OpenJDK users to also install OpenJFX), however I wouldn't really like to do that since neither Terasology or FacadeServer are JavaFX applications, and besides the OpenJDK issue I'd prefer not to depend on a full blown GUI framework like JavaFX just to have an implementation of a pair.

There are 3 alternative solutions I can think of:
* Use a `Pair` implementation from one of the libraries already required for Terasology (which we'll always have on the classpath). I think [Guava has one](https://docs.atlassian.com/fugue/1.2.0/apidocs/com/atlassian/fugue/Pair.html), but I'm not sure if we use that Guava extension (as said I'd prefer not to add a new dependency for a pair). [Apache Commons](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/tuple/Pair.html) should have one as well (but I don't remember if we also use commons for Terasology? you may want to have a look at the engine's Gradle files)
* Build your own `Pair` class - should be quite simple (using Java generics)
* Refactor the code to use custom classes instead of pairs

I'd personally prefer 2 or 3 since 1 would lead to problems in case an external library is removed from the engine's dependencies (for example, from a quick search of the main repo seems like Guava is included but not widely used)

src/main/java/org/terasology/web/serverAdminManagement/AdminPermissionManager.java
```diff
@@ -0,0 +1,206 @@
...
+
+    @In
+    private NetworkSystem networkSystem;
+    @In
+    private PermissionManager permissionManager;
+
+    private final Path adminPermissionsFilePath;
+    private final Type typeOfServerAdminPermissions = new TypeToken<Set<Pair<String, Map<PermissionType, Boolean>>>>() {
+    }.getType();
+    private Set<Pair<String, Map<PermissionType, Boolean>>> serverAdminPermissions;
```
gianluca-nitti: Is there a particular reason for using `Map<PermissionType, Boolean>` instead of simply a `Set<PermissionType>`?

src/main/java/org/terasology/web/serverAdminManagement/AdminPermissionManager.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+        Pair<String, Map<PermissionType, Boolean>> adminPermission = getPermissionsOfAdmin(id);
+        serverAdminPermissions.remove(adminPermission);
+        onListChanged.run();
+    }
+
+    public Pair<String, Map<PermissionType, Boolean>> getPermissionsOfAdmin(String id) {
+        for (Pair<String, Map<PermissionType, Boolean>> adminPermission : serverAdminPermissions) {
+            if (adminPermission.getKey().equals(id)) {
+                return adminPermission;
```
gianluca-nitti: In my understanding you could only return the right part (or value) of the pair, which has the actual information about permissions - the left part (or key) is the id, which the caller of this method already knows since it passes it as argument, correct?

src/main/java/org/terasology/web/client/ClientSecurityInfo.java
```diff
@@ -15,21 +15,31 @@
...
+import org.terasology.web.serverAdminManagement.PermissionType;
+
+import java.util.List;
+
 public class ClientSecurityInfo {
 
     private boolean isAuthenticated;
-    private boolean hasAdminPermission;
+    private boolean isAdmin;
+    private List<PermissionType> clientPermissions;
```
gianluca-nitti: I think a `Set` would be more suited since the order is not relevant, correctly? If you use a set and also replace `Map<Something, Boolean>` with `Set<Something>` as suggested in the other comments (if it's possible) you may be able to do various simplifications of the code since you won't need to do conversions between collection types (by iterating over the values) anymore

(my fault since I suggested you `List`, but I was just thinking of a generic collection, sorry)

gianluca-nitti (commented):
Thanks for your patience @Inei1, and please excuse me if I'm being a bit pedantic with the reviews - I'm confident it's almost ready to merge :slightly_smiling_face:, still needs a bit of cleanup in my opinion.

As you can see by the comments
* it would be better to avoid using classes from the JavaFX package (`Pair` in this case)
* the other tips should be useful to simplify the code where possible (without affecting functionality, obviously)

src/main/java/org/terasology/web/serverAdminManagement/AdminPermissionManager.java
Inei1: I need to store the permission true/false values somewhere in `serverAdminPermissions`, so I thought it would be easier to understand and code by using a map.

src/main/java/org/terasology/web/serverAdminManagement/IdPermissionPair.java
```diff
@@ -0,0 +1,67 @@
...
+package org.terasology.web.serverAdminManagement;
+
+import java.util.Objects;
+
+/**
+ * This class is a simple container that follows the pair pattern.
+ * @param <T> the first part of the pair.
+ * @param <U> the second part of the pair.
+ */
+public class IdPermissionPair<T, U> {
```
gianluca-nitti: I think this class is attempting to mix two goals. What should it actually represent?
- If it should model a generic pair, like the one from JavaFX, using the T and U type parameters is correct but I think the names of the two fields shouldn't refer to a specific usage (`id` and `permissions`), but should be something like `first` and `second` (like [C++'s pair](http://www.cplusplus.com/reference/utility/pair/)), or `key` and `value` (like [JavaFX's pair](https://docs.oracle.com/javase/8/javafx/api/javafx/util/Pair.html) as you already know), or `left` and `right`, etc. In this case, I'd also move the class to a different package (maybe there is an `utils`/`utilities` package somewhere, or you may create one at the root of the project) since it can also be used for other purposes which aren't related to server admin management 
- If you want it to only model an *id-permissions pair*, then I'd keep these names for the fields but I wouldn't use generics - an id is always a `String` and a set of permissions is always a `Map<PermissionType, Boolean>`. In this case I think the class is already in the correct package, since it's only used for admin management purposes

gianluca-nitti (approved):

gianluca-nitti:
Okay, merged - thank you @Inei1 for the work.

I did some testing and the code mostly works as expected. An annoying fact is that console-related permissions are not automatically updated for users connected via the web client (including the one which requested the modification) - for the changes to take effect, you have to connect with the game client.
After a bit of investigation I found out that this is due to the headless clients (i.e. users connected via the web client) only being registered to the engine's `EntityManager` and not to the `NetworkSystem`, thus `AdminPermissionManager` (which [queries the `NetworkSystem`](https://github.com/MovingBlocks/FacadeServer/pull/17/files#diff-4fcb1a7627af44f60e8d0a4bd3868dffR117)) can't find them, and no `onConnected` event is fired when an headless client connects, so [this code in `ConsoleResource`](https://github.com/MovingBlocks/FacadeServer/pull/17/files#diff-9dc022b4f06e33c300fabaf4a1f5915eR50) isn't executed for headless clients (only for normal clients), meaning that refreshing the frontend in the browser and reconnecting doesn't help.

This is mostly due to a poor design choice I made when I wrote `HeadlessClient` and other classes, since I thought that just registering the entity was enough - but after getting more familiar with the engine I'm quite sure the proper way is to register them to the `NetworkSystem`.

~I'm going to open an issue~ Just opened issue #19 (mostly as a note to self, but anyone interested is welcome to help) as a reminder to fix this.

PR #18, opened 2018-06-19
Prepare FacadeServer for packaging in Omega distribution

Inei1:
Add a few changes to the jar file, make new launch scripts, and change ServerMain to give information on keystore files.

This needs to be merged before I finalize the changes to the omega distribution, as it adds the scripts I need to copy over.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,6 +66,13 @@ dependencies {
     testCompile group: 'org.mockito', name: 'mockito-core', version: '2.7.22'
 }
 
+jar {
+    manifest {
+        attributes('Main-Class': mainClassName)
+        attributes('Class-Path': configurations.runtime.collect { it.getName() }.join(' '))
+    }
+}
+
 // Copied from PC facade
 // By delaying this task to doLast (the << bit) we don't get the headless server dir set up unless actually wanting it
 // TODO: This is not the Gradle Way. Needs more declared output-fu to determine up-to-date instead of the if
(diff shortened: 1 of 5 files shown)


gianluca-nitti:
Changes for finding the keystore and building the jar file look fine.

I don't understand/remember however how the `libs/FacadeServer-1.0.0-SNAPSHOT.jar` file which the scripts should launch has to be built (i.e. which Gradle command I need to run) - can you please provide a short recap of the required steps (i.e. which commands one should run to build the FacadeServer binary distribution and which files have to be copied into the Omega distribution zip)? Thanks

Inei1:
So here is what I think should work in the omega distribution:
1. Jenkins builds the jar (using `gradlew jar` if I understand it correctly).
2. [distros.gradle](https://github.com/Terasology/Index/blob/master/distros/distros.gradle) copies the jar and all its dependencies into the libs folder.
[Here](https://pastebin.com/DcvqUb0J) is the modified distros.gradle script I used to make this work.
3. (This one needs this PR to be done first to get the launch scripts) distros.gradle copies in the files from the launchScripts directory. Also, we can't do this the same way the PC facade does it because it copies in the scripts beforehand to make the base Terasology.zip.

You can run `gradlew distPC` in a local version of [the omega workspace](https://github.com/Terasology/Index/tree/master/distros/omega) with my modified script to copy in the server facade jar and all its dependencies. After that, if you manually copy in the launch scripts, it should work fine on Windows (and possibly on Linux/Mac, but it's untested there).

gianluca-nitti:
Okay, thank you for the explanation - merging.

I managed to build an Omega zip which includes FacadeServer using your instructions.

It may be worth noticing that when downloading FacadeServer from Artifactory, Gradle also downloaded an `engine-2.0.1-20180619.023039-5.jar` - which in my understanding shouldn't be needed since the engine should already be present in the Terasology.zip manually copied in the build environment, correct? But this is probably to be discussed in another PR (in the Index repository probably).

Issue #19, opened 2018-06-22
HeadlessClient instances should be registered to the engine's NetworkSystem (not just to the EntityManager)


gianluca-nitti:
What the title says. Right now clients which connects via the HTTP/WebSocket service exposed by FacadeServer are registered to the engine's `EntitySystem`, so that they can run console commands - however, it would be better to register them to the `NetworkSystem` (which job is to keep track of all the connected clients).
This should allow `ComponentSystem`s to receive event notifications (e.g. `onConnected`) related to headless clients, like they do for regular clients.
Besides improving the code structure (with `NetworkSystem` handling all kinds of clients, including "special" ones), right now an use case where this would improve usability is in the features intruduced in #17 (see [this comment](https://github.com/MovingBlocks/FacadeServer/pull/17#issuecomment-399584624)).

PR #21, opened 2018-06-24
Add world map resource

Inei1:
Adds a world map resource to the REST API.

To test, connect to the game using a normal client (so that the blocks will be loaded, otherwise they all show up as engine:unloaded) and query the worldMap resource.


src/main/java/org/terasology/web/resources/worldMap/WorldMapMetadata.java
```diff
@@ -0,0 +1,83 @@
...
+                        logger.error("unable to get the texture for block " + blockUrn.toString());
+                    }
+                }
+                blocks.add(blockUrn.toString());
+            }
+        }
+    }
+
+    private int getAverageRGB(BufferedImage bufferedImage) {
+        long average = 0;
```
gianluca-nitti: As said on Slack I think this may not be correct since the results [`bufferedImage.getRGB`](https://docs.oracle.com/javase/7/docs/api/java/awt/image/BufferedImage.html#getRGB(int,%20int)) are 32-bit numbers in the format `AAAAAAAARRRRRRRRGGGGGGGGBBBBBBBB` where As are the bits for the alpha (transparency) channel, Rs for red and so on. So if you sum two of these numbers, the sum of a channel may overflow in the bits of the one on its right (e.g. blue in green) and you'll get incorrect results.

What I'd do is averaging the single channels, e.g. declare an int for R, G and B, and sum to each of them in the inner for loop. The you could return an instance of `org.terasology.rendering.nui.Color`, I think it has a constructor which takes the values of every individual channel.

src/main/java/org/terasology/web/resources/worldMap/WorldMapMetadata.java
```diff
@@ -0,0 +1,83 @@
...
+        blockLut = new HashMap<>();
+        this.getWorldMapData();
+    }
+
+    private void getWorldMapData() {
+        for (int x = center.getX() - mapBlockWidth / 2; x < mapBlockWidth / 2; ++x) {
+            for (int z = center.getZ() - mapBlockLength / 2; z < mapBlockLength / 2; ++z) {
+                Block block = worldProvider.getBlock(x, center.getY(), z);
+                ResourceUrn blockUrn = block.getURI().getBlockFamilyDefinitionUrn();
+                if (blockLut.get(blockUrn.toString()) == null) {
```
gianluca-nitti: I think you are doing a good thing by caching the results of `getAverageRGB`, which is quite computationally intensive, in `blockLut` and only computing the average texture color for a certain block if you haven't already, however I think it's possible to do better: we could also cache them across requests. In my understanding, right now every requests builds a new instance of `WorldMapMetadata`, which builds it's own `blockLut`; however, this is not necessary - the textures, and thus the colors, are always the same.

I'd suggest something like making `blockLut` static, thus not sending it with each frame, and adding another resource to access it separately (which the client will only request once). This would also make it easier to migrate to the textured version later if we decide to do so, since we will only need to send the texture instead of the color (once we find the proper way to serialize it).

src/main/java/org/terasology/web/resources/worldMap/WorldMapMetadata.java
```diff
@@ -0,0 +1,105 @@
...
+        blockLut = new HashMap<>();
+        this.getWorldMapData();
+    }
+
+    WorldMapMetadata(WorldProvider worldProvider, Vector3i topLeft, int mapBlockWidth, int mapBlockLength) {
+        this.topLeft = topLeft;
+        this.mapBlockWidth = mapBlockWidth;
+        this.mapBlockLength = mapBlockLength;
+        this.worldProvider = worldProvider;
+        blocks = new ArrayList<>();
```
gianluca-nitti: (Probably already known, but I'll just drop a reminder here) Note that you also need to initialize the inner arrays as you do in the other test constructor, otherwise you'll get NPEs when calling `add` on line 81

src/main/java/org/terasology/web/resources/worldMap/WorldMapResource.java
```diff
@@ -0,0 +1,44 @@
...
+    private WorldProvider worldProvider;
+
+    @Override
+    protected ResourceMethod<Void, WorldMapMetadata> getGetMethod(ResourcePath path) throws ResourceAccessException {
+        return createParameterlessMethod(path, ClientSecurityRequirements.PUBLIC, Void.class,
+                (data, client) -> new WorldMapMetadata(worldProvider));
+    }
+
+    @Override
+    protected ResourceMethod<WorldMapInput, WorldMapMetadata> getPutMethod(ResourcePath path) throws ResourceAccessException {
```
gianluca-nitti: Semantically (see [HTTP/1.1 Method Definitions](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)), only GET would be the proper choice - PUT is meant to modify the resource, and the map isn't actually being modified here.

However I can see why you made this choice (I suppose) - there is no way to parametrize a GET request in our resource system. So I'd say it's fine to keep it this way until we update the system to support it. After all in plain HTTP (opposed to this custom wrapper/abstraction layer for HTTP and WebSocket) a request made with the GET method *can* be parametrized, usually with the [query string](https://en.wikipedia.org/wiki/Query_string).

gianluca-nitti (commented):
Tested together with MovingBlocks/FacadeServer-frontend#8 and worked fine :slightly_smiling_face: as long as a player is online.

If this condition isn't satisfied, a grid of unloaded blocks is returned and the map is shown black. I think - but am not 100% sure - that implementing #19 could fix this, i.e. by registering the "fake"/headless player(s) to the NetworkSystem, in my understanding, the engine should "see" them and load the appropriate chunks. 

However I think we may also need to move the player's in-game position within the area for the requested map to force the engine to load the involved part of the world.

src/main/java/org/terasology/web/resources/worldMap/WorldMapResource.java
```diff
@@ -0,0 +1,200 @@
...
+        loadChunks(center, mapBlockWidth, mapBlockLength);
+
+        for (int x = (int) Math.floor((double) center.getX() - mapBlockWidth / 2); x < (int) Math.ceil((double) mapBlockWidth / 2 + center.getX()); ++x) {
+            for (int z = (int) Math.floor((double) center.getZ() - mapBlockLength / 2); z < (int) Math.ceil((double) mapBlockLength / 2 + center.getZ()); ++z) {
+                /*synchronized (this) {
+                    while (worldProvider.getBlock(x, blockY, z).getURI().toString().equals("engine:unloaded")) {
+                        try {
+                            wait(1000);
+                        } catch (InterruptedException e) {
+                            e.printStackTrace();
```
Inei1: We still need a way to give time for the chunks to load after chunk loading is triggered. You can see my first attempt here, which didn't have good enough performance.

src/main/java/org/terasology/web/resources/worldMap/WorldMapResource.java
```diff
@@ -0,0 +1,232 @@
...
+                    BufferedImage blockImage = Assets.get(blockUrn, BlockTile.class).get().getImage();
+                    colors.get(x - (center.getX() - mapBlockWidth / 2)).add(new Color(getColorOfTexture(blockImage, blockY)));
+                } else {
+                    logger.warn("cannot find texture of block " + blockUrn.toString());
+                    colors.get(x - (center.getX() - mapBlockWidth / 2)).add(Color.BLACK);
+                }
+            }
+        }
+
+        int colorSize = (int) (1 / Math.floor(Math.log(mapBlockWidth * mapBlockLength)) * colorSizeMultiplier);
```
gianluca-nitti, line 122: Two things:
1) It took me a while to understand this is the length of the side of a square representing a block in the image, I had to get to line 117 to get it - and in my understanding it hasn't much to do with colors, maybe this could be renamed to something more meaningful like `blockSize` or `blockSide`?
2) I don't fully understand the formula, personally I think I'd have just used a constant size (like `colorSizeMultiplier`). If I understand correctly `mapBlockWidth * mapBlockLength` is the area measured in blocks, but I don't understand why you compute it's logarithm and take it's inverse, could you please explain what's the reason?

src/main/java/org/terasology/web/resources/worldMap/WorldMapResource.java
```diff
@@ -0,0 +1,232 @@
...
+     * @param mapBlockLength the length of the map.
+     */
+    private void loadChunks(Vector3i center, int mapBlockWidth, int mapBlockLength) {
+        final int maximumVerticalChunks = 8;
+        LocationComponent locationComponent = new LocationComponent();
+        locationComponent.setWorldPosition(center.toVector3f());
+        RelevanceRegionComponent relevanceRegionComponent = new RelevanceRegionComponent();
+        relevanceRegionComponent.distance = new Vector3i(((int) Math.ceil((double) mapBlockWidth / ChunkConstants.SIZE_X) * 2) + 2, maximumVerticalChunks,
+                ((int) Math.ceil((double) mapBlockLength / ChunkConstants.SIZE_Z) * 2) + 2);
+        entityManager.create(locationComponent, relevanceRegionComponent);
```
gianluca-nitti: In my understanding the entity created here is never removed, so I think this can leak memory which in the long run (after a lot of requests) could cause problems (waste of memory) (does it sond reasonable to you?). I think a solution could be not ignoring the result returned here and instead return in to the caller of `loadChunks`; then on line 87 we could save the `EntityRef` in a variable, and somehow destroy it when the chunks are no longer necessary (e.g. around line 110 I think). I think the proper way should be calling [`EntityRef.destroy()`](https://github.com/MovingBlocks/Terasology/blob/971c78fd322ce81e9d067b6cea0bc882a6584222/engine/src/main/java/org/terasology/entitySystem/entity/EntityRef.java#L57) but someone with better knowledge of the entity system may be able to confirm or provide more information.

gianluca-nitti (commented):
Tested this out together with FacadeServer-frontend#8 and works fine, thanks for the work.

I have a couple of questions about the `colorSize` variable and a concern about leaking memory by not deleting the entities which are created on every request to load the chunks, but as said in the comments it should be quite simple to fix.

So almost ready to be merged, I'd say.

src/main/java/org/terasology/web/resources/worldMap/WorldMapResource.java
Inei1: I think renaming the variable to `imageBlockSize` would work.

The formula is used because otherwise, the image displayed on the frontend would be too large for a 250x250 map, and too small for a 20x20 map. The formula works to normalize the size of the images. It's the best way I could find to do it, but if you have any suggestions for a better solution then I would be glad to hear it.

gianluca-nitti: Ah okay thanks for the explanation. It works fine and I can't think of any better way right now, so I agree for keeping your formula

PR #22, opened 2018-07-14
Add Resources for the Blacklist and Whitelist 

Inei1:
Adds two new resources for the blacklist and whitelist. These features were implemented in the engine by me before GSoC started (see https://github.com/MovingBlocks/Terasology/pull/3332).

To test, use the frontend and look at the new blacklist/whitelist management button on the user management tab. The only thing that seems to work is the GET requests, the other ones display an error when used. I'm not sure why they don't work because I tried to copy the existing serverAdmins resource as closely as I could.


gianluca-nitti:
Tested and works as expected. Thanks!

PR #25, opened 2018-07-28
Add More Unit Tests

Inei1:
Adds new unit tests to the backend.

### DIFF
--- a/src/test/java/org/terasology/web/resources/console/ConsoleResourceTest.java
+++ b/src/test/java/org/terasology/web/resources/console/ConsoleResourceTest.java
@@ -15,31 +15,67 @@
  */
 package org.terasology.web.resources.console;
 
+import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.terasology.context.Context;
 import org.terasology.context.internal.ContextImpl;
 import org.terasology.entitySystem.entity.EntityRef;
+
 import org.terasology.logic.console.Console;
+import org.terasology.logic.console.ConsoleImpl;
 import org.terasology.logic.console.Message;
 import org.terasology.logic.console.MessageEvent;
+import org.terasology.logic.console.commandSystem.MethodCommand;
+import org.terasology.logic.console.commands.ServerCommands;
 import org.terasology.network.Client;
+import org.terasology.network.ClientComponent;
+import org.terasology.network.NetworkMode;
+import org.terasology.network.NetworkSystem;
 import org.terasology.registry.InjectionHelper;
 import org.terasology.web.resources.base.ResourceAccessException;
 import org.terasology.web.resources.base.ResourceObserver;
 import org.terasology.web.resources.base.ResourcePath;
 
+import static org.junit.Assert.assertNotNull;
 import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
 
 public class ConsoleResourceTest {
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    private Console consoleMock;
+    private Context context;
+    private Client client;
+    private EntityRef clientEntityMock;
+    private ConsoleResource consoleResource;
+
+    @Before
+    public void setupConsole() {
+        consoleMock = mock(Console.class);
+        context = new ContextImpl();
+        context.put(Console.class, consoleMock);
+        consoleResource = new ConsoleResource();
+        InjectionHelper.inject(consoleResource, context);
+
+        client = mock(Client.class);
+        clientEntityMock = mock(EntityRef.class);
+        when(client.getEntity()).thenReturn(clientEntityMock);
+    }
+
+    @Test
+    public void testGetConsoleCommands() throws ResourceAccessException {
+        setupConsole();
+        assertNotNull(consoleResource.getGetMethod(ResourcePath.createEmpty()).perform(null, client));
+    }
 
     @Test
     @SuppressWarnings("unchecked")
     public void testMessageNotification() {
-        ConsoleResource consoleResource = new ConsoleResource();
         ResourceObserver observer = mock(ResourceObserver.class);
         MessageEvent testEvent = mock(MessageEvent.class);
         Message message = new Message("testMessage");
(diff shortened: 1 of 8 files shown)


src/test/java/org/terasology/web/serverAdminManagement/AdminPermissionManagerTest.java
```diff
@@ -0,0 +1,134 @@
...
+        Context context = new ContextImpl();
+        context.put(PermissionManager.class, permissionManagerMock);
+        AdminPermissionManager adminPermissionManager = createPermissionListWithTwoAdmins();
+        InjectionHelper.inject(adminPermissionManager, context);
+
+        // The permission manager mock does not correctly add the permissions
+        adminPermissionManager.updateAdminConsolePermissions("first", clientEntityRefMock);
+
+        // Therefore, this is false
+        assertTrue(permissionManagerMock.hasPermission(clientEntityRefMock, PermissionManager.CHEAT_PERMISSION));
```
Inei1: Any hints on why this doesn't work? It seems like either the mock isn't working correctly or I'm using the permission system incorrectly.

src/test/java/org/terasology/web/resources/console/ConsoleResourceTest.java
```diff
@@ -15,31 +15,67 @@
...
 import org.terasology.web.resources.base.ResourceObserver;
 import org.terasology.web.resources.base.ResourcePath;
 
+import static org.junit.Assert.assertNotNull;
 import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
gianluca-nitti: Minor note - our [coding conventions](https://github.com/MovingBlocks/Terasology/wiki/Code-Conventions#others) suggest avoiding star imports.
If you use IntelliJ there is a setting to automate this: `File > Settings`, then `Editor > Code Style > Java`, `Imports` tab; there are two fields `Class count to use import with '*'` and `Names count to use static import with '*'` - I think if you set both of them to a large number (like, say, 50) the IDE should use the proper style when adding imports.

src/test/java/org/terasology/web/resources/console/ConsoleResourceTest.java
```diff
@@ -15,31 +15,67 @@
...
+        InjectionHelper.inject(consoleResource, context);
+
+        client = mock(Client.class);
+        clientEntityMock = mock(EntityRef.class);
+        when(client.getEntity()).thenReturn(clientEntityMock);
+    }
+
+    @Test
+    public void testGetConsoleCommands() throws ResourceAccessException {
+        setupConsole();
```
gianluca-nitti: Since `setupConsole` is annotated with `@Before`, it shouldn't be necessary to call it manually. JUnit takes care of automatically running it before every test in this class (see [reference](http://junit.sourceforge.net/javadoc/org/junit/Before.html)).

FYI there is also a [`@BeforeClass`](http://junit.sourceforge.net/javadoc/org/junit/BeforeClass.html) annotation which can be useful in certain situations; it sets up the annotated method to run just once before running the various test in the class.

src/test/java/org/terasology/web/resources/console/ConsoleResourceTest.java
```diff
@@ -50,19 +86,44 @@ public void testMessageNotification() {
         verify(observer).onEvent(argThat(ResourcePath::isEmpty), eq(message), eq(clientEntity));
     }
 
+    @Test
+    public void testCommandExecutionFails() throws ResourceAccessException {
+        setupConsole();
```
gianluca-nitti: Same here - not needed...

src/test/java/org/terasology/web/resources/console/ConsoleResourceTest.java
```diff
@@ -50,19 +86,44 @@ public void testMessageNotification() {
...
+        ServerCommands serverCommands = new ServerCommands();
+        MethodCommand.registerAvailable(serverCommands, consoleWithCommandsSpy, context);
+
+        consoleResource.getPostMethod(ResourcePath.createEmpty()).perform("save", client);
+        verify(consoleWithCommandsSpy).execute("save", clientEntityMock);
+    }
+
+    @Test
+    public void testHelpCommand() throws ResourceAccessException {
+        setupConsole();
```
gianluca-nitti: ..and here.

src/test/java/org/terasology/web/resources/console/ConsoleResourceTest.java
```diff
@@ -50,19 +86,44 @@ public void testMessageNotification() {
         verify(observer).onEvent(argThat(ResourcePath::isEmpty), eq(message), eq(clientEntity));
     }
 
+    @Test
+    public void testCommandExecutionFails() throws ResourceAccessException {
```
gianluca-nitti: If I understand it correctly, it should fail because the command doesn't exist (i.e. [this] should be the line throwing the exception; is it correct?). It took me a while to realize, so I'd make it explicit in the method name, like `testUnexistingCommandExecutionFails`.

src/test/java/org/terasology/web/resources/console/ConsoleResourceTest.java
```diff
@@ -50,19 +86,44 @@ public void testMessageNotification() {
...
+    }
+
+    @Test
+    public void testHelpCommand() throws ResourceAccessException {
+        setupConsole();
+
+        ResourceObserver resourceObserverMock = mock(ResourceObserver.class);
+        consoleResource.setObserver(resourceObserverMock);
+        consoleResource.getPostMethod(ResourcePath.createEmpty()).perform("help", client);
+        verify(consoleMock, never()).execute(anyString(), eq(clientEntityMock));
```
gianluca-nitti: I agree with verifying that no real command is actually executed by the console.

However unless I'm missing something, `resourceObserverMock` is never used. Maybe the idea was to verify that it received a notification with the help message? I think just verifying it received a notification (`any()` matchers could be useful) should be enough.

src/test/java/org/terasology/web/resources/worldMap/WorldMapResourceTest.java
```diff
@@ -0,0 +1,87 @@
...
+        Context context = new ContextImpl();
+        context.put(WorldProvider.class, worldProviderMock);
+        context.put(EntityManager.class, entityManagerMock);
+        InjectionHelper.inject(worldMapResource, context);
+    }
+
+    @Test
+    public void testColorOfTexture() {
+        // predetermined to be the correct color through testing
+        final int arbitraryImageColorToTest = 3947580;
```
gianluca-nitti: Just an idea - would it work if we compute the average between the 4 colours here, something like
```
final int r = (Color.BLACK.getR() + Color.GREEN.getR() + Color.BLUE.getR() + Color.RED.getR()) / 4;
//same for g and b
final int arbitraryImageColorToTest = new Color(r, g, b).getRGB();
```
In my understanding it should give the correct result since the 4 5x5 squares all have the same surface, so the weight of each color is the same, correct?
This way we would avoid the hardcoded value while still testing that `getColorOfTexture` is analyzing all the pixels.

src/test/java/org/terasology/web/resources/worldMap/WorldMapResourceTest.java
```diff
@@ -0,0 +1,87 @@
...
+        assertEquals(arbitraryImageColorToTest, worldMapResource.getColorOfTexture(getTestImage(), 75));
+    }
+
+    @Test
+    public void testImageConvertsToBase64String() {
+        BufferedImage testImage = getTestImage();
+        assertNotNull(worldMapResource.convertImageToBase64String(testImage));
```
gianluca-nitti: Unfortunately this is somewhat tricky to test... an idea could be converting the other way round - from base64 to byte array and building a BufferedImage from it and checking for example that the size (width and height) is the same

src/test/java/org/terasology/web/resources/console/ConsoleResourceTest.java
Inei1: `resourceObserverMock` is used because otherwise the `notifyEvent` method in `AbstractObservableResource` throws an NPE, since it tries to access the observer.

Issue #26, opened 2018-08-04
Better handling of command line arguments


gianluca-nitti:
A nice but still missing feature would be supporting the same command line arguments supported by the regular PC facade, like for example `-overrideDefaultConfig`.

Right now they are mostly managed [by the PC facade itself](https://github.com/gianluca-nitti/Terasology/blob/develop/facades/PC/src/main/java/org/terasology/engine/Terasology.java), so a first step could be considering if it's the case to migrate some code from that class to the engine package to be able to reuse it for other facades (like Server) and avoid duplicating code.

gianluca-nitti:
Closing as handling of the most relevant arguments has been implemented by @Inei1 in #29.

PR #29, opened 2018-09-03
Handle more command line arguments

Inei1:
Solves #26. There are three command line arguments in the PC facade that work with a headless server added by this PR. More testing is needed for the `-overrideDefaultConfig` argument, but the others work perfectly. Most of the new code is duplicated from the PC facade.

### DIFF
--- a/src/main/java/org/terasology/web/ServerMain.java
+++ b/src/main/java/org/terasology/web/ServerMain.java
@@ -45,6 +45,8 @@ import org.glassfish.jersey.server.mvc.freemarker.FreemarkerMvcFeature;
 import org.glassfish.jersey.servlet.ServletContainer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.terasology.config.Config;
+import org.terasology.config.SystemConfig;
 import org.terasology.engine.LoggingContext;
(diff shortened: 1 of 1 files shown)


gianluca-nitti:
Thanks - tested and works fine, also `-overrideDefaultConfig`. My quick test consisted in
* preparing this command line - through IntelliJ's "Program Arguments" field - `-homedir=terasology-server -overrideDefaultConfig=override.cfg`, the same as the default headless run configuration without `-headless` which is obvious for the server facade
* deleting the `terasology-server` directory to start from a clean environment, without config and old saves
* modifying `override.cfg` by adding `"defaultSeed": "someTestSeed"` to the `"worldGeneration"` section
* launching FacadeServer with the command line mentioned above
* waiting for it to save
* opening `terasology-server/saves/New World/manifest.json`, and as expected the `"seed"` entry was`"someTestSeed"` instead of the default `"blockmania"`

gianluca-nitti:
Small GitHub tip: when a PR fixes an issue, like in this case, you can use [one of the keywords](https://help.github.com/articles/closing-issues-using-keywords/), like `fixes` instead of `solves`, and when the PR is merged the issue is automagically closed :slightly_smiling_face:

